PR #2, opened 2018-08-23
Make toolchain work under Archlinux

hperl:
The following changes were necessary to use the toolchain under Archlinux.

First, the autodetection script needs a case for Archlinux. Arch works with the current Ubuntu binaries, so not much work was necessary there.

Second, Archlinux links python -> python3 by default. This is not really according to the PEP, but that PEP came after the change and not much can be done here. Except, since this is really a Python 2 script, we should use `python2` instead of assuming the symlink python -> python2.


siddharthab:
Thanks!

PR #10, opened 2019-04-13
Bazel 0.24.0 compatibility

### LINKED ISSUE #9: Bazel 0.24 compatibility
Using this toolchain with Bazel 0.24.0 fails, the `cc_toolchain` rule doesn't have a `dynamic_runtime_libs` nor a `static_runtime_libs` attribute anymore.

Kernald:
Removes `dynamic_runtime_libs` and `static_runtime_libs` from the `conditional_cc_toolchain` macro, as they were set to `:empty` and doesn't exist anymore in Bazel 0.24.0. This fixes #9.

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -97,10 +97,8 @@ def conditional_cc_toolchain(name, cpu, darwin, absolute_paths = False):
             all_files = ":empty",
             compiler_files = ":empty",
             dwp_files = ":empty",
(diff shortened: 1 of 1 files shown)


siddharthab:
Thank you! The test for Ubuntu 14.04 is failing because it is still using bazel 0.19.1. I have removed support for 14.04 in a6bc74685a9e21df1063b2cd9c8d3c7dcc1fcd3d, so this change should be good to go.

PR #17, opened 2019-08-06
Add support for debian testing (this is the "head" debian release).

mhines01:
this doesn't contain a version id and causes the python script to trace.

Example from debian:

PRETTY_NAME=“Debian GNU/Linux buster/sid”
NAME=“Debian GNU/Linux”
ID=debian
HOME_URL=“https://www.debian.org/”
SUPPORT_URL=“https://www.debian.org/support”
BUG_REPORT_URL=“https://bugs.debian.org/”

Based on this commit  it looks like the VERSION_ID maybe dropped
in future versions of debian?

https://metadata.ftp-master.debian.org/changelogs//main/b/base-files/base-files_11_changelog

### DIFF
--- a/toolchain/tools/llvm_release_name.py
+++ b/toolchain/tools/llvm_release_name.py
@@ -68,6 +68,8 @@ def _linux(llvm_version):
         os_name = "linux-gnu-ubuntu-18.04"
     elif distname in ["arch", "ubuntu"] or (distname == "linuxmint" and version.startswith("18")):
         os_name = "linux-gnu-ubuntu-16.04"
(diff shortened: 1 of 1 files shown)


siddharthab:
Thank you for your contribution. Can you move these lines to just below the lines for version 9. 

Also do you know if version 9 should be using ubuntu 16.04 release, or ubuntu 18.04? If you think the latter, can you combine the two `elif` clauses?

mhines01:
debian 9 is based on ubuntu 16.04 it should be correct as it was

siddharthab:
Thank you!

PR #18, opened 2019-08-19
Adds manjaro support

WilliamIzzo83:
the 'unknown distribution error' happens on [manjaro](https://manjaro.org) distro.
This fixes the issue.


siddharthab:
Thanks!

PR #28, opened 2019-10-23
toolchain/BUILD.tpl: add missing ':ar' dependency

jmcmullan:
When linking a static library, the following error could occur:

```
ERROR: BUILD.bazel:123:1: Linking of rule '//foo:libfoo' failed (Exit 1)
external/io_bazel/src/main/tools/process-wrapper-legacy.cc:58: "execvp(external/llvm_toolchain/bin/llvm-ar, ...)": No such file or directory
```

This was due to `:linker_components` missing a dependency on the `:ar` target.
This change add `:ar` to the `:linker_components` target.

Signed-off-by: Jason S. McMullan <jason.mcmullan@uber.com>

### DIFF
--- a/toolchain/BUILD.tpl
+++ b/toolchain/BUILD.tpl
@@ -194,6 +194,7 @@ filegroup(
     srcs = [
         ":clang",
         ":ld",
+        ":ar",
         ":lib",
         ":sysroot_components",
     ],


siddharthab:
Jason, is it possible for you to either update your bazel version to 1.0+, or to go to a commit before 2200d53?

I am thinking of a good way to message to people that going forward, this project is going to have a strict requirement on bazel 1.0 or up so we start relying on their shipped configurations. I guess I will have to load skylib to check bazel version when they register the toolchain using the provided method.

siddharthab:
Let's merge this now, and I will do a new release when we actually start using features from bazel 1.0 onwards.

PR #29, opened 2019-10-24
GNU libstdc++

johnmarkwayve:
Switch from LLVM libc++ to GNU libstdc++ 

This is necessary for compatibility with ROS2 and we should be able to undo it if ROS2 is retired.


johnmarkwayve:
Apologies, this was accidentally submitted to the wrong repo

PR #30, opened 2019-10-25
Fix being unable to fetch LLVM 9.0 on Mac OS

sabarabc:
The name of the LLVM tar changed from "clang+llvm-{llvm_version}-x86_64-apple-darwin.tar.xz" to "clang+llvm-{llvm_version}-x86_64-darwin-apple.tar.xz" with LLVM 9.0.0.


siddharthab:
Thank you for the PR. I will submit another commit to switch our default tested version to be 9.0.0.

PR #36, opened 2020-03-03
Workaround to NO-OP on Windows

flaub:


siddharthab:
Thanks!

I wonder what error you were seeing. We do use bazel on some Windows machines to build go binaries, but we don't see an error during repo initialization.

flaub:
This is where it was blowing up:
https://github.com/grailbio/bazel-toolchain/blob/master/toolchain/internal/llvm_distributions.bzl#L153

siddharthab:
Sorry for the delay. I forgot about this PR.

What you are saying makes sense. Even though we build Windows binaries, we cross-compile them on Linux where a valid toolchain could still be downloaded. Hopefully the situation will be better once we have proper cross-compilation support.

PR #37, opened 2020-03-08
toolchain: on Fedora + CentOS, use the Ubuntu 18.04 clang

bpowers:
I was hoping it would avoid errors I see on a Fedora host like:

```
external/llvm_toolchain/bin/clang: /lib64/libtinfo.so.5: no version information available (required by external/llvm_toolchain/bin/clang)
```

But it doesn't; however Fedora releases are much closer to 18.04
anyway, so I think this would be a good change anyway.

### DIFF
--- a/tests/scripts/fedora_test.sh
+++ b/tests/scripts/fedora_test.sh
@@ -29,7 +29,7 @@ set -exuo pipefail
 
 # Install dependencies
 dnf install -qy dnf-plugins-core
-dnf install -qy python gcc
+dnf install -qy python gcc ncurses-compat-libs
 
 # Run tests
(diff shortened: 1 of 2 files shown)


siddharthab:
Thank you for your PR.

Can you also fix the test in fedora_test.sh? You can run that script directly to run the test.

Most likely you will have to install libtinfo5.so in the docker image in that test. I can also look into it if it's not clear.

bpowers:
@siddharthab I've fixed that Fedora test!  I'm not sure why the mac build is failing -- I think it is unrelated to this PR?

siddharthab:
Thank you!

PR #39, opened 2020-03-29
Add LLVM 10.0.0 support

per-gron:
From LLVM 10.0.0, `https://releases.llvm.org/` is no longer a valid URL prefix for LLVM prebuilt binaries; they have moved the hosting to github. Unfortunately for this script, they also changed the directory naming convention. I changed the script to not change how the `llvm_mirror` attribute is used, but this requires that the official LLVM mirror URLs need to be treated slightly differently from `llvm_mirror`.


per-gron:
The  (macos-latest, run_external_tests.sh)  test looks broken, it fails with 403 when downloading bazel releases, which seems unrelated to this PR.

siddharthab:
Thank you so much for the PR! I will try to update our default to 10.0.0 and see what dependencies it needs on the various linux distros.

PR #41, opened 2020-04-14
Update centos 7 to use sles toolchain as Ubuntu uses a versino of gli…

jsharpe:
…bc that is too new for centos 7


siddharthab:
Thanks!

PR #46, opened 2020-05-14
sysroot: support passing '/' as the sysroot

adunham-stripe:
If we have a container that contains libraries/header files/etc., we may want to use the LLVM toolchain but the platform-wide sysroot. This PR changes the sysroot detection to not crash if we pass a sysroot of `"/"`.

### DIFF
--- a/toolchain/internal/sysroot.bzl
+++ b/toolchain/internal/sysroot.bzl
@@ -41,7 +41,10 @@ def sysroot_path(rctx):
     if not sysroot:
         return (_default_sysroot(rctx), None)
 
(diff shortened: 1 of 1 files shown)


adunham-stripe:
Error appears unrelated:
```
2020/05/14 19:14:20 could not resolve the version 'latest' to an actual version number: could not get releases from github.com/bazelbuild/bazel: could not download list of Bazel releases from github.com/bazelbuild: unexpected status code while reading https://api.github.com/repos/bazelbuild/bazel/releases: 403
```

siddharthab:
Thank you!

PR #49, opened 2020-06-17
Export files under bin/ and lib/

kornholi:
This allows someone, for example, to depend on `lib/libclang.so`.

A good portion of these files are already implicitly exported, but that behaviour will be changing soon. See https://github.com/bazelbuild/bazel/issues/10225.


siddharthab:
Thanks!

siddharthab:
Now included in tag 0.5.6.

PR #54, opened 2020-06-21
Change the archive fetched on Archlinux

### LINKED ISSUE #53: Archlinux and LLVM 10.0.0
Trying to use LLVM 10.0.0 on Archlinux with the latest version of this toolchain (0.5.6) raises this error:

```
Unknown LLVM release: clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz
```

For some reason, [here](https://github.com/grailbio/bazel-toolchain/blob/master/toolchain/tools/llvm_release_name.py#L77) Archlinux is assumed to be a Ubuntu 16.04 equivalent indeed. It worked fine with LLVM 8.0.0 and 9.0.0, but LLVM doesn't provide prebuilt archives for Ubuntu 16.04 on x86_64 anymore for LLVM 10.0.0 (list [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-10.0.0)).

Kernald:
LLVM 10.0.0 hasn't been published for Ubuntu 16.04 (and likely won't),
and given the rolling-release nature of Archlinux, it makes more sense
to stick to the most recent OS targeted by LLVM.

This fixes #53

### DIFF
--- a/toolchain/tools/llvm_release_name.py
+++ b/toolchain/tools/llvm_release_name.py
@@ -91,7 +91,7 @@ def _linux(llvm_version):
     elif (distname == "ubuntu" and version.startswith("20")) or (distname == "pop" and version.startswith("20")):
         # use ubuntu 18.04 clang LLVM release for ubuntu 20.04
         os_name = "linux-gnu-ubuntu-18.04"
-    elif distname in ["arch", "ubuntu", "manjaro"] or (distname == "linuxmint" and version.startswith("18")):
+    elif distname in ["ubuntu", "manjaro"] or (distname == "linuxmint" and version.startswith("18")):
         os_name = "linux-gnu-ubuntu-16.04"
     elif distname == "debian" and (version is None or int(version) == 10):
         os_name = "linux-gnu-ubuntu-18.04"
@@ -108,6 +108,12 @@ def _linux(llvm_version):
         os_name = "linux-gnu-ubuntu-18.04"
     elif distname == "amzn" and major_llvm_version >= 7:
         os_name = "linux-gnu-ubuntu-18.04"
+    elif distname == "arch" and major_llvm_version >= 11:
+        os_name = "linux-gnu-ubuntu-20.04"
+    elif distname == "arch" and major_llvm_version >= 10:
+        os_name = "linux-gnu-ubuntu-18.04"
+    elif distname == "arch" and major_llvm_version >= 7:
+        os_name = "linux-gnu-ubuntu-16.04"
     else:
         sys.exit("Unsupported linux distribution and version: %s, %s" % (distname, version))
 


siddharthab:
Hi! Thanks for the PR. Will it be possible for you to fix the container based test for archlinux? If not, I can try.

Kernald:
I'm looking at it right now, I'm a little bit confused. It looks like LLVM now depends on ncurses, and more specifically ncurses 5 - but didn't with Ubuntu 16.04 builds? Archlinux ships with ncurses 6, hence the error. There's a compat package on AUR, but installing a package from AUR on CI would be a bit of a pain...

siddharthab:
Does llvm 10 also link ncurses-5? Right now, our default version for the tests is llvm 9.

Kernald:
Yeah, both of them do. I sadly haven't found a pre-built archive (at least pre-built by LLVM) linked to ncurses 6 (that's the case of the version provided by Archlinux packages). It happens to work on my machine as I have a ncurses5 compat package from AUR installed, but as I mentioned installing a package from AUR on CI doesn't sound like a great idea. I'm not sure what's the best approach here...

siddharthab:
Hmm... Now that I remember, I kept the default version in these tests to be llvm 8, because of this issue. I wonder if we are not getting llvm 8 archive in the tests.

https://github.com/grailbio/bazel-toolchain/blob/7e7c7cf1f965f348861085183d79b6a241764390/WORKSPACE#L27

Kernald:
Ha, good catch. I just downloaded LLVM 8.0.0 for both Ubuntu 16.04 and 18.04, and indeed clang isn't linked to the same librairies in both cases:

```
➜ ldd clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04/bin/clang | grep libtinfo

➜ ldd clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04/bin/clang | grep libtinfo
        libtinfo.so.5 => /usr/lib/libtinfo.so.5 (0x00007f4250645000)
```

I guess `libncurses5` wasn't installed out of the box on Ubuntu 16.04 but is on Ubuntu 18.04. I guess one solution would be to stick to the Ubuntu 16.04 release for LLVM<10.0.0, but switch to the 18.04 one for >= 10.0.0. Not a great solution, but I don't think it would change anything in practice and would fix the issue on CI.

Kernald:
Short update on this (finally):
 - Works for llvm 8 to 10 (starting from 10, it gets a bit tricky to test on CI as `libtinfo.so.5` is only provided by AUR packages - namely `ncurses5-compat-libs`
 - llvm 11 seems to be working, but raises a bunch of warnings: 
    ```
    external/llvm_toolchain/bin/clang: /usr/lib/libtinfo.so.6: no version information available (required by 
    external/llvm_toolchain/bin/clang)
    ```
    `libtinfo.so.6` is the one provided by `core/ncurses`, not an AUR package. I'm not sure what to do about that.

siddharthab:
Sorry about my extremely prolonged absence. This looks good. If there are any further changes needed, I can make them myself. Thank you for checking back on this.

PR #61, opened 2020-09-14
Add netrc and auth pattern support for downloading LLVM distribution archives

dylanplecki:
This diff adds support for using netrc-based authentication or optional auth patterns for download LLVM distribution archive files from preconfigured or manually configured URLs.

The netrc file, if not specified, will be automatically loaded from well-known locations on Unix/Windows, which mirrors the functionality of `http_archive`.

### DIFF
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc", "use_netrc")
+
 # If a new LLVM version is missing from this list, please add the shasum here
 # and send a PR on github. To compute the shasum block, you can use the script
 # utils/llvm_checksums.sh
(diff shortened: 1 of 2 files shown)


dylanplecki:
It looks like tests are failing outside of this change, `bazel build //tests/...` does work successfully after this change locally.

toolchain/internal/llvm_distributions.bzl
```diff
@@ -131,6 +133,31 @@ def _python(rctx):
...
+    Based on:
+    https://github.com/bazelbuild/bazel/blob/793964e8e4268629d82fabbd08bf1a7718afa301/tools/build_defs/repo/http.bzl#L42
+    """
+    if ctx.attr.netrc:
+        netrc = read_netrc(ctx, ctx.attr.netrc)
+        return use_netrc(netrc, urls, ctx.attr.auth_patterns)
+
+    if "HOME" in ctx.os.environ and not ctx.os.name.startswith("windows"):
+        netrcfile = "%s/.netrc" % (ctx.os.environ["HOME"])
+        if ctx.execute(["test", "-f", netrcfile]).return_code == 0:
```
siddharthab, line 149: Can you leave a comment as to why `ctx.path(netrcfile)` does not work here, like you do for USERPROFILE? I think the code in the bazel repo is of very old origin and hence the use of `test`.

siddharthab:
Apologies for the delay. I will merge this as-is, and resolve the review comments in a later commit.

PR #62, opened 2020-09-18
Support Pop!_OS 20.04

yagehu:


yagehu:
@siddharthab Can you take a quick look? This simply adds support for Pop OS 20.04. It also fixes a small undefined name bug.

siddharthab:
Sorry, this got neglected by me.

Issue #14, opened 2019-06-25
Plain "python" is not available on many modern distributions


refi64:
Out-of-the-box, most distros are shipping Python 3 nowadays, which means that no "python" binary is available without installing Python 2.

siddharthab:
Thank you for the report. Can you give me a few examples of such distros?

My understanding is that while [PEP 394](https://www.python.org/dev/peps/pep-0394/) is active, OS distros should continue to include a `python` command.

refi64:
The gotcha is here: 
> for the time being, all distributions should ensure that python, **if installed**, refers to the same target as python2

Since Python 2 is reaching EOL next year, many distros already don't have it installed by default, and others are trying to begin to drop it altogether.

In particular, both Fedora and RHEL 8 use Python 3 by default.

siddharthab:
Understood. I will update the shebang line after testing with our supported OS distros.

PR #65, opened 2020-10-15
python -> python3

jez:
Ubuntu 20.04 does not include python on the PATH by default.

    ❯ which -a python
    python not found

    ❯ which -a python3
    /usr/bin/python3
    /bin/python3

Recent Python distributions all install the python3 binary.

This tool uses one python script (the one in this PR), and
it obviously assumes Python 3, because it uses `print(...)`
without `from __future__ import print_function`.

Would you accept a patch to hardcode `python3` in the shebang?
I'd really like to leave Python 2 behind, and there is too
much software in the world that assumes `python` means `python2`.

### DIFF
--- a/toolchain/tools/llvm_release_name.py
+++ b/toolchain/tools/llvm_release_name.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Copyright 2018 The Bazel Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");


jez:
This PEP is relevant:

https://www.python.org/dev/peps/pep-0394/

> When packaging third party Python scripts, distributors are encouraged to change less specific shebangs to more specific ones.

siddharthab:
Thanks!

For this change, we actually stopped relying on the shebang line in resolving #14. So I am a little surprised that your change actually fixes your issue because it should effectively be a no-op.

Perhaps we can also just delete the shebang line now because we don't rely on it. I will merge this PR as-is for now.

jez:
Gotcha sorry about that. I am actually using an ages old fork of this repo
at work, and I’d like to not be, so I started pulling commits out to
upstream them.

It’s very possible that if I had been able to run on HEAD then the issue is
already fixed.

PR #66, opened 2020-10-25
Work even if the repository is loaded under a different name

jpieper:
With this change, you can load the repository under a name other than "@com_grail_bazel_toolchain", for instance if you are using a forked version or just want to give it another name.

I'm not aware of any problems this causes.  Is it something you'd be willing to consider?


siddharthab:
Sure thing. Apologies for the delay.

PR #67, opened 2020-10-25
Factor the OS name normalization into one place

jpieper:
Is this something you'd be willing to accept?  The only functional change is that you get an error slightly earlier on Windows, but otherwise it just de-duplicates the logic that is used to create the OS short name.


siddharthab:
Thanks. It deleted the Windows bypass but I can restore it in a follow-up commit.

PR #72, opened 2020-12-19
Support Debian releases with non-integer VERSION_ID

johnsheu:
The systemd specification for the /etc/os-release file, where we pull the system version information from, specifies that the VERSION_ID field may contain characters in the range 0–9, a–z, ".", "_" and "-".  This commit fixes parsing for VERSION_ID fields that contain more than just the numeric digits 0–9.


siddharthab:
Apologies for the delay. Now merged.

PR #77, opened 2021-03-16
Add support for 11.0.1 and 11.1.0

djmarcin:
Created with `utils/llvm_checksums.sh -g -v 11.0.1` and `utils/llvm_checksums.sh -g -v 11.1.0`

### DIFF
--- a/tests/scripts/ubuntu_20_04_test.sh
+++ b/tests/scripts/ubuntu_20_04_test.sh
@@ -30,7 +30,7 @@ set -exuo pipefail
 # Common setup
 export DEBIAN_FRONTEND=noninteractive
 apt-get -qq update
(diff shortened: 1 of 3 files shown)


djmarcin:
Auto-detection leaves something to be desired, as there is no 20.04 package for these versions of LLVM -- they appear to have been bumped to 20.10, and the 20.10 packages don't work on 20.04. There are no 18.04 packages, but the 16.04 packages appear to work on 20.04.

siddharthab:
Thank you for the PR and my apologies in the delay in getting to it.

11.1.0 was already added in a different PR, so I used this PR to add 11.0.1. The auto-detection logic has also been improved for these two special releases.

PR #79, opened 2021-04-28
Add support for aarch64 on user-provided URLs

gabrielrussoc:
Currently we can only provide URLs for `linux` and `darwin`, regardless of the architecture. We recently started to support aarch64 and there was no way to provide a custom llvm toolchain for it.

This needs some polishing but I'm putting it up for review to gather initial feedback. Also in case something like this is already on the works.

The code was written by @ryan-johnson-databricks


toolchain/BUILD.tpl
```diff
@@ -73,7 +75,7 @@ toolchain(
 )
 
 toolchain(
-    name = "cc-toolchain-linux",
+    name = "cc-toolchain-amd64-linux",
```
brian-brt, line 78: How about sticking with k8 here? I don't see amd64 used in any other starlark, and keeping the pattern the same will help me find all the pieces when I go looking for them.

toolchain/cc_toolchain_config.bzl.tpl
```diff
@@ -49,31 +53,40 @@ def _impl(ctx):
         target_system_name = "x86_64-apple-macosx"
     elif (ctx.attr.cpu == "k8"):
         target_system_name = "x86_64-unknown-linux-gnu"
+    elif (ctx.attr.cpu == "aarch"):
+        target_system_name = "aarch-unknown-linux-gnu"
```
brian-brt, line 57: Typo? I've never figured out what this value affects though, seems like nothing important.

toolchain/cc_toolchain_config.bzl.tpl
```diff
@@ -503,7 +519,7 @@ def _impl(ctx):
         "%{toolchain_path_prefix}lib/clang/%{llvm_version}/include",
         "%{toolchain_path_prefix}lib64/clang/%{llvm_version}/include",
     ]
-    if (ctx.attr.cpu == "k8"):
+    if (ctx.attr.cpu in ("k8", "aarch64")):
         cxx_builtin_include_directories += [
             "%{sysroot_prefix}/include",
             "%{sysroot_prefix}/usr/include",
```
brian-brt, line 525: 3 lines down needs updating.

brian-brt (commented):
I've got something very similar I was looking into upstreaming, when I noticed this PR :) I took a look, it's pretty similar to my approach (nothing exciting there). Noticed a few little things though.

gabrielrussoc:
@brian-brt I'm happy to drop this for something better. This was basically a hack we did in order to unblock and I opened the PR for visibility.

siddharthab:
Thanks for the PR. I don't think I have permissions to update your PR branch, so I will merge this as commit independent of the PR.

siddharthab:
Merged with 85022a2d.

PR #82, opened 2021-05-17
Put amzn as a known distro but treat it like sles 12.4

brianbentson:
AMZN was never added  to the `_known_distros` list however it was added to the switch statement in https://github.com/grailbio/bazel-toolchain/pull/48

### DIFF
--- a/toolchain/tools/llvm_release_name.py
+++ b/toolchain/tools/llvm_release_name.py
@@ -71,12 +71,6 @@ def _linux(llvm_version):
     # If you find this mapping wrong, please send a Pull Request on Github.
     if arch in ["aarch64", "armv7a", "mips", "mipsel"]:
         os_name = "linux-gnu"
-    # amzn linux was defaulting to sles11.3 because of a typo, this is likely the closest distro based on the ID_LIKE field
-    elif distname == "amzn":
-        if major_llvm_version >= 11:
-            os_name = "linux-sles12.4"
-        else:
-            os_name = "linux-sles11.3"
     elif distname == "freebsd":
         os_name = "unknown-freebsd-%s" % version
     elif distname == "suse":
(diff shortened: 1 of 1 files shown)


toolchain/tools/llvm_release_name.py
```diff
@@ -71,6 +71,9 @@ def _linux(llvm_version):
     # If you find this mapping wrong, please send a Pull Request on Github.
     if arch in ["aarch64", "armv7a", "mips", "mipsel"]:
         os_name = "linux-gnu"
+    # amzn linux was defaulting to sles11.3 because of a typo, this is likely the closest distro
+    elif distname == "amzn":
+        os_name = "linux-sles12.4"
```
siddharthab, line 76: You need to version gate this. Not all LLVM releases since LLVM 6 have distros for SLES 12.4.

brianbentson: for sure just added

toolchain/tools/llvm_release_name.py
```diff
@@ -71,6 +71,9 @@ def _linux(llvm_version):
     # If you find this mapping wrong, please send a Pull Request on Github.
     if arch in ["aarch64", "armv7a", "mips", "mipsel"]:
         os_name = "linux-gnu"
+    # amzn linux was defaulting to sles11.3 because of a typo, this is likely the closest distro
```
siddharthab, line 74: So, LLVM releases are actually available for SLES 11.3 until LLVM 10 (that's why the ID_LIKE logic was working until now for llvm version < 11). What you want here is:

```
elif distname == "amzn":
  if major_llvm_version >= 11:
    os_name = "linux-sles12.4"
  else:
    os_name = "linux-sles11.3"
```

siddharthab: Also, the comment can be changed to say:
```
amzn linux has its ID_LIKE field set to SLES, but we want to explicitly control the SLES version.
```

brianbentson: > Also, the comment can be changed to say:
> 
> ```
> amzn linux has its ID_LIKE field set to SLES, but we want to explicitly control the SLES version.
> ```

It actually only has [centos, RHEL, fedora] in the ID_LIKE field. SLES is just close to those as well.

siddharthab: OK. Let me update the comment before merging.

Issue #87, opened 2021-09-01
Debian 11 Support in LLVM toolchain


JoshVarga:
Debian 11 (which was officially released August 14 2021) is not explicitly supported currently, but seems that using the following would work correctly.

os_name = "linux-gnu-ubuntu-18.04"

The check for OS version is here:
https://github.com/grailbio/bazel-toolchain/blob/master/toolchain/tools/llvm_release_name.py#L95-L100

Would you accept a pull request for this change?

siddharthab:
This was picked up in f54297f4 when I was fixing the container test failures. The container test for debian now uses Debian 11. Thanks.

PR #89, opened 2021-09-02
Adding Debian 11 as a supported OS

JoshVarga:
I added Debian 11 as a supported OS as mentioned in this issue: https://github.com/grailbio/bazel-toolchain/issues/87

No significant changes that I have been able to find have been made to Debian 11 that would prevent this library from working the same as Debian 10.


siddharthab:
Thank you for the PR. I handled this in f54297f4, when I was looking at the container test failures. The container test for debian now uses Debian 11.

PR #103, opened 2021-09-24
Moving lpthread and ldl from link_flags to link_libs

Anomander:
Putting lpthread into link_flags takes it to the front of link command,
causing linker to remove unused symbols.
This change moves -lpthread and -ldl from link_flags to link_libs which
correctly puts them at the tail.

### DIFF
--- a/tests/stdlib.cc
+++ b/tests/stdlib.cc
@@ -15,26 +15,3 @@
 #include <iostream>
 
 void hello() { std::cout << "Hello World!" << std::endl; }
-
-#if defined(__linux__) && defined(_GNU_SOURCE)
-#include <dlfcn.h>
-#include <assert.h>
-
-// Checks that pthread symbols are loadable.
-//
-// This test verifies that dynamically linked libraries which
-// rely on pthread symbols can still access them.
-// Incorrect order of linking dependencies may remove unused
-// symbols and break shared libraries that are linked later.
-void test_pthread_symbols() {
-   // Find symbol
-   void* symbol = dlsym(RTLD_NEXT, "pthread_getspecific");
-   // Check that there is no error
-   assert(dlerror() == nullptr && symbol != nullptr);
-}
-
-#else//defined(__linux__) && defined(_GNU_SOURCE)
-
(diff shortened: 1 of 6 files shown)


Anomander:
`bazel build //...` and `bazel test //...` pass

toolchain/cc_toolchain_config.bzl
```diff
@@ -126,6 +126,11 @@ def cc_toolchain_config(
         "-lm",
         "-no-canonical-prefixes",
     ]
+    link_libs = [
```
siddharthab, line 129: Can we just declare `link_libs` here to be `[]` and then add the actual values to when we actually do use libunwind:
```
                "-l:libunwind.a",
                # Compiler runtime features.
                "-rtlib=compiler-rt",
            ])
            link_libs.extend([
                # To support libunwind.
                "-lpthread",
                "-ldl",
            ])

**Review by siddharthab — CHANGES_REQUESTED** [review]
Should we also use `link_libs` for libc++ or libstdc++?

**Comment by siddharthab** [comment]
Will it also be possible to add a test case that exposes the issue? Maybe just modify stdlib.cc to use pthreads, or create a new pthreads.cc.

**Comment by Anomander** [comment]
Let me try this.

File: toolchain/cc_toolchain_config.bzl (hunk 1 above)
Comment (Anomander) [reply toolchain/cc_toolchain_config.bzl]: > Should we also use link_libs for libc++ or libstdc++?

I'm not sure. In our use-cases, it doesn't really matter, but there may be others.

**Comment by siddharthab** [comment]
Thanks.

**Comment by Anomander** [comment]
Extracted the pthread verification into its own test, since it doesn't work with fully-statically linked binaries.

**Review by siddharthab — APPROVED** [review]

File: tests/stdlib.cc [hunk 2]
```diff
@@ -15,3 +15,26 @@
 #include <iostream>
 
 void hello() { std::cout << "Hello World!" << std::endl; }
+
```
siddharthab, line 18: Looks good. Just one last thing. Changes from this file need to be deleted.

PR #105, opened 2021-09-25
Symlink '/usr/bin/ld' to llvm toolchain directory on Mac

dozzman:
Mac systems still use the local `ld` executable since `lld.ld` is still
experimental. Some applications, e.g. rustc when compiling rust, expect
to find `ld` within the same directory as the compiler and so when
compiling rust we run into an issue where the linker is not found.

The first immediate solution is to use the rust rule's
`extra_rustc_flags` to specify the linker explicitly (via
`-Clinker=/usr/bin/ld`. This works for some crates but fails for crates
using a build script, since the extra flags are not propogated to rustc
invocations within the exec configuration.

Another solution attempt was to link `ld` into the toolchain
configuration repository instead of directly in the llvm repository,
such that `ld` becomes a sibling file to `cc_wrapper.sh`. Unfortunately
rustc was still unable to find the linker with this configuration.

The final fully working solution is to link `ld` into the bin folder of
the llvm repository. This should be fine as LLVM's own linker exists
under `ld.lld` and symlinking the local linker will not clobber any
files.


siddharthab:
Thank you for the PR.

We used to do this when GoCompile had the same problem. I removed these symlinks in 6e4288 because Go rules did not need it anymore. Will you be able to add some tests for Rust in the run_external_tests.sh script so we can track compatibility?

rrbutani:
I'm still a little confused about why/how `rustc` seems to be searching the `bin` directory of the LLVM repository instead of the `bin` dir where `cc_wrapper.sh` lives but that doesn't need to block this PR.

I added a test to `run_external_tests.sh` that runs `git2-rs`'s tests.

rrbutani:
@siddharthab Adding `rules_rust` causes our migration test to fail for [`--incompatible_no_rule_outputs_param=true`](https://github.com/bazelbuild/bazel/issues/7977).

I can think of three ways to get around this:
  1) add `--incompatible_no_rule_outputs_param=false` to `test_args` in `run_tests.sh`
     - this seems like a bad precedent: we're taking on the full set of migration problems of our test dependencies
     - in this specific case I don't think it's a big deal (we're pretty unlikely to grow any `outputs` in our rules as a toolchain, I think) but still
  2) move `rules_rust` and its tests to another workspace
     - I don't think we can move _all_ the tests to another workspace (or at least I don't think this would help test migrations) since some of the migrations affect more than just things in the load phase, I think (on the other hand we currently only really expose a repo rule anyways)
     - the downside here is duplication (though there wouldn't be that much of it)
  3) gate the `rules_rust` bits in `WORKSPACE` on `$TEST_MIGRATION` by wrapping those bits in a repo rule
    - this gets around the duplication problem but is maybe a little too much magic? not sure

I'm partial to the third option but whatever you think is best works.

---

As a sidenote: we don't actually seem to run `run_external_tests.sh` in `migration.yml` yet it has checks for `$TEST_MIGRATION`; is this intentional?

siddharthab:
I would not worry too much about the migration failure. Just adding an exception in run_tests.sh (option 1) seems fine. If possible, file an issue with rules_rust and mention the issue in comments before our exception.

For run_external_tests.sh, we can remove the block that adds migration check flags. I don't think I put too much thought into it when I added it there.

WORKSPACE
```diff
@@ -138,3 +138,97 @@ http_archive(
...
+
+load("@rules_rust//rust:repositories.bzl", "rust_repositories")
+
+rust_repositories(
+    version = "1.55.0",
+    edition = "2018",
+)
+
+# We're using `git2` as our Rust test because it links against C code
+# (`libgit2-sys`), has tests, and is non-trivial but not _massive_.
```
siddharthab, line 160: Can you also comment why any of the tests in rules_rust itself are not sufficient. For rules_go, we just run their unit tests for cgo. So would be nice to note why we are doing things differently for rules_rust.

tests/foreign/git2-rs-cargo-toml.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/Cargo.toml b/Cargo.toml
```
siddharthab, line 1: The foreign directory is for rules_foreign_cc. Can we keep this patch in a new directory called rust?

siddharthab (changes requested):

WORKSPACE
rrbutani: Some of the `rules_rust` tests break when run from another workspace (as in when `rules_rust` is a dependency and not the main workspace).

The main reason though, is that I wanted the test to have a build script that tries to build some C code with the `cc` crate. `rules_rust` doesn't have a test that does this afaik.

I can add a comment explaining this.

dozzman:
The force push was a mistake -- thought I accidentally pushed some commits, didn't realise there was some other work done on the branch. Have reverted that.

tests/foreign/git2-rs-cargo-toml.patch
dozzman: Have moved this.

dozzman:
> I would not worry too much about the migration failure. Just adding an exception in run_tests.sh (option 1) seems fine. If possible, file an issue with rules_rust and mention the issue in comments before our exception.
> 
> For run_external_tests.sh, we can remove the block that adds migration check flags. I don't think I put too much thought into it when I added it there.

Added the appropriate flag to `run_tests.sh` and `run_external_tests.sh`. The other scripts looked like they didn't need amending.

WORKSPACE
dozzman: So the `ld` error message rears its head even when compiling a very simple build script. My previous force push had a simple crate with a build script + test program which consumed the build script's output. Running the test with `--@rules_rust//:extra_rustc_flags=-Clinker=/usr/bin/ld` gave only an error when building the build script. Re-enabling the symlink gets the build to complete 100%.

Although I think it's wise to have tests which use the `cc` crate as its more representative of what folks may be using a build script for.

dozzman:
Was sure I saw a message requesting to remove the `$TEST_MIGRATION` blocks but since I can no longer find it I will leave it.

WORKSPACE
siddharthab: Thank you both. I am not a Rust user so will trust you in your judgement on what is best to test here. An in-line code comment summarizing this thread would be nice and then we can merge this.

rrbutani: @dozzman oh I misunderstood your initial message; I thought you were saying `-Clinker` wasn't being passed down to `cc` compiler invocations, not to the `rustc` invocation that builds the build script (for `cargo` at least this is [intended behavior](https://github.com/rust-lang/cargo/issues/4423) I think)

Regardless, I agree; I think it's unlikely we manage to do something in this repo that breaks `cc` called from build scripts but not other parts of `rules_rust` but I'll feel better knowing that we're testing it :).

@siddharthab I updated the comment in `WORKSPACE` to mention the bits about `cc` and the `rules_rust` tests not working when run from another workspace.

rrbutani:
> Was sure I saw a message requesting to remove the `$TEST_MIGRATION` blocks but since I can no longer find it I will leave it.

[This](https://github.com/grailbio/bazel-toolchain/pull/105#issuecomment-927203378) was the message I think. I've removed that bit from `run_external_tests.sh`.

rrbutani:
I did a little more digging; `rustc` calls `cc_wrapper.sh` as the "linker" which then calls `clang` which then tries to shell out to `ld`.

`clang` can't find `ld` when `ld` is next to `cc_wrapper.sh` but _can_ when `ld` is next to `clang` itself (as in, in the LLVM repo's bin folder) which makes sense.

I'm still a little confused why we only seem to run into this when `cc_wrapper.sh` is called from `rustc`; nothing about the `cc_wrapper.sh` call `rustc` generates looks different. Maybe it's something odd about `$PWD` in that case.

Regardless, I think this PR is good to go :shipit:.

siddharthab:
> I did a little more digging; `rustc` calls `cc_wrapper.sh` as the "linker" which then calls `clang` which then tries to shell out to `ld`. `rustc` itself [_does_](https://github.com/rust-lang/rust/blob/308dffd25cb55bbb4a1fbee9822cf82c6a5d012d/compiler/rustc_session/src/session.rs#L804-L813) do some searching for the linker it calls but that's not what's failing here.

That's because in that function, PATH is set to `/bin`, but the linker on macOS is in `/usr/bin`.

siddharthab (approved):
I will merge this as-is and then maybe commit some more changes to clean up later.

rrbutani:
> > I did a little more digging; `rustc` calls `cc_wrapper.sh` as the "linker" which then calls `clang` which then tries to shell out to `ld`. `rustc` itself [_does_](https://github.com/rust-lang/rust/blob/308dffd25cb55bbb4a1fbee9822cf82c6a5d012d/compiler/rustc_session/src/session.rs#L804-L813) do some searching for the linker it calls but that's not what's failing here.
> 
> That's because in that function, PATH is set to `/bin`, but the linker on macOS is in `/usr/bin`.

Sorry, that snippet is a little misleading; `$PATH` isn't being set to `/bin` but `/bin` is being added to a list of search directories. Regardless, `rustc` doesn't error there; `cc_wrapper.sh` is specified by `rules_rust` as the linker so `rustc` just calls `cc_wrapper.sh` (without doing any searching) which _then_ errors when `clang` can't find `ld`.

The actual `rustc` invocation looks something like this on my machine:
```sh
bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/rules_rust/util/process_wrapper/process_wrapper --subst 'pwd=${pwd}' -- external/rust_darwin_x86_64/bin/rustc external/crates__log__0_4_14/build.rs '--crate-name=log_build_script_script_' '--crate-type=bin' '--error-format=human' '--codegen=metadata=' '--out-dir=bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14' '--codegen=extra-filename=' '--codegen=opt-level=3' '--codegen=debuginfo=0' '--remap-path-prefix=${pwd}=.' '--emit=dep-info,link' '--color=always' '--target=x86_64-apple-darwin' -L external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib '--cap-lints=allow' '--codegen=linker=external/llvm_toolchain/bin/cc_wrapper.sh' --codegen 'link-args=-mlinker-version=450 --target=x86_64-apple-macosx -lm -no-canonical-prefixes -headerpad_max_install_names -undefined dynamic_lookup -lc++ -lc++abi --sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk'
```

Which then leads to a cc_wrapper.sh invocation that looks like this:
```sh
"external/llvm_toolchain/bin/cc_wrapper.sh" "-m64" "-arch" "x86_64" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.0.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.1.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.2.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.3.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.4.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.5.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.6.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.7.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.log_build_script_script_.11355ed9-cgu.8.rcgu.o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_.5esazdxjfc5wg0qb.rcgu.o" "-L" "external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib" "-L" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libstd-35b14ca5e62c7d66.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libpanic_unwind-329b6ae10a542427.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libobject-09e9356fd4907aa1.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libaddr2line-f077c0a492da74c3.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libgimli-01a2a2a1730ab072.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libstd_detect-736ff6c48d99c951.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_demangle-f6b7985fe24842c3.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libhashbrown-6ccb889196a16125.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_std_workspace_alloc-2bda93c117f030a0.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libunwind-bcbd09b25c64fcac.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcfg_if-ef7e97011650176d.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/liblibc-104a032674bfeab8.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/liballoc-ba3aef5f5e0e8573.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/librustc_std_workspace_core-9ee1c33f7a829255.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcore-20d0dacb579b8ca8.rlib" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-2fee5b6fdc3293e9.rlib" "-lSystem" "-lresolv" "-lc" "-lm" "-liconv" "-L" "/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/lib" "-o" "bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/crates__log__0_4_14/log_build_script_script_" "-Wl,-dead_strip" "-nodefaultlibs" "-mlinker-version=450" "--target=x86_64-apple-macosx" "-lm" "-no-canonical-prefixes" "-headerpad_max_install_names" "-undefined" "dynamic_lookup" "-lc++" "-lc++abi" "--sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"
```

I can try to run the `rustc` `cc_wrapper.sh` invocation under `dtruss` and compare it to a "regular" `cc_wrapper.sh` invocation from a `rules_cc` target to figure out where the discrepancy is coming from later.

rrbutani:
@siddharthab what other cleanup needs to happen?

siddharthab:
> Sorry, that snippet is a little misleading; `$PATH` isn't being set to `/bin` but `/bin` is being added to a list of search directories.

You can verify my claim by adding a `echo $PATH >> /tmp/foo` in osx_cc_wrapper.sh, and then running the test.

rrbutani:
@siddharthab you're absolutely right; good catch

I'm still fairly convinced it's not `rustc` that's doing this; I think it's actually [this](https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/cargo/cargo_build_script.bzl#L66-L68) from `rules_rust` (it'd explain why we only see the problem for build scripts too).

PR #106, opened 2021-09-26
Ensure /usr/bin is in PATH for macOS

siddharthab:
This is an alternative fix to what was included in #105.

Thanks for @rrbutani for discovering the underlying issue.

### DIFF
--- a/toolchain/osx_cc_wrapper.sh.tpl
+++ b/toolchain/osx_cc_wrapper.sh.tpl
@@ -62,6 +62,26 @@ for i in "$@"; do
     fi
 done
 
+# On macOS, we use ld as the linker for single-platform builds (i.e., when not
+# cross-compiling). Some applications may remove /usr/bin from PATH before
+# calling this script, which would make /usr/bin/ld unreachable.  For example,
+# rules_rust does not set PATH (unless the user explicitly sets PATH in env
+# through attributes) [1] when calling rustc, and rustc does not replace an
+# unset PATH with a reasonable default either ([2], [3]), which results in CC
+# being called with PATH={sysroot}/{rust_lib}/bin. Note that rules_cc [4] and
+# rules_go [5] do ensure that /usr/bin is in PATH.
+# [1]: https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/cargo/cargo_build_script.bzl#L66-L68
+# [2]: https://github.com/rust-lang/rust/blob/1c03f0d0ba4fee54b7aa458f4d3ad989d8bf7b34/compiler/rustc_session/src/session.rs#L804-L813
+# [3]: https://github.com/rust-lang/rust/blob/1c03f0d0ba4fee54b7aa458f4d3ad989d8bf7b34/compiler/rustc_codegen_ssa/src/back/link.rs#L640-L645
+# [4]: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java;l=529;drc=72caead7b428fd50164079956ec368fc54a9567c
+# [5]: https://github.com/bazelbuild/rules_go/blob/63dfd99403076331fef0775d52a8039d502d4115/go/private/context.bzl#L434
+# Let's restore /usr/bin to PATH in such cases. Note that /usr/bin is not a
+# writeable directory on macOS even with sudo privileges, so it should be safe
+# to add it to PATH even when the application wants to use a very strict PATH.
+if [[ ":${PATH}:" != *":/usr/bin:"* ]]; then
+  PATH="${PATH}:/usr/bin"
+fi
+
 # Call the C++ compiler.
 if [[ -f %{toolchain_path_prefix}bin/clang ]]; then
   %{toolchain_path_prefix}bin/clang "$@"
--- a/README.md
+++ b/README.md
@@ -178,7 +178,8 @@ the / read-only mount. Note that this will make your builds non-hermetic.
 
 #### Compatibility
 
-The toolchain is tested to work with `rules_go` and `rules_foreign_cc`.
+The toolchain is tested to work with `rules_go`, `rules_rust`, and
+`rules_foreign_cc`.
 
 #### Accessing tools
 
--- a/toolchain/internal/repo.bzl
+++ b/toolchain/internal/repo.bzl
@@ -35,7 +35,7 @@ def llvm_repo_impl(rctx):
 
     _download_llvm_preconfigured(rctx)
 
-    # darwin may use local 'ld' so symlink it to bin directory to help
-    # other programs locate it when called directly (e.g. rustc)
-    if os == "darwin":
-        rctx.symlink("/usr/bin/ld", "bin/ld")
+    # We try to avoid patches to the downloaded repo so that it is easier for
+    # users to bring their own LLVM distribution through `http_archive`. If we
+    # do want to make changes, then we should do it through a patch file, and
+    # document it for users of toolchain_roots attribute.


siddharthab:
@dozzman Would you prefer this solution instead of the symlink?

rrbutani:
@siddharthab it'll be interesting to see what CI says but I don't think this actually works; `rustc` isn't actually removing anything from `$PATH` afaik (the snippet I linked to is where it searches for tools when the linker isn't explicitly specified)

rrbutani:
CI [passed](https://github.com/grailbio/bazel-toolchain/actions/runs/1274366505) on the version before the comment change/force push 😃.

rrbutani:
As [mentioned](https://github.com/grailbio/bazel-toolchain/pull/105#issuecomment-927226188) on the other issue, I think it's actually [this bit](https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/cargo/cargo_build_script.bzl#L66-L68) in `rules_rust` that's responsible for dropping `/usr/bin` from `$PATH`.

`rules_rust` [actually](https://github.com/bazelbuild/rules_rust/blob/52e3ba4c99692b5354faff0fb1aa62ddf884024c/rust/private/rustc.bzl#L232-L236) uses [`cc_common.get_environment_variables`](https://docs.bazel.build/versions/main/skylark/lib/cc_common.html#get_environment_variables) to get the env vars to use in its invocation; I wonder if there's a way to get that function to keep `/usr/bin` in the `$PATH`.

Regardless, I think this is a better fix in #105 so that, as you mentioned, users can provide their own LLVM repos without needing patches or accidentally breaking macOS host support.

toolchain/osx_cc_wrapper.sh.tpl
```diff
@@ -62,6 +62,18 @@ for i in "$@"; do
...
 done
 
+# On macOS, we use ld as the linker for single-platform builds. Some
+# applications may remove /usr/bin from PATH before calling this script, which
+# would make /usr/bin/ld unreachable.
+# For example, rules_rust sets PATH to be /bin.
+# https://github.com/rust-lang/rust/blob/308dffd25cb55bbb4a1fbee9822cf82c6a5d012d/compiler/rustc_session/src/session.rs#L804-L813
+# Let's restore /usr/bin to PATH in such cases. Note that /usr/bin is not a
+# writeable directory on macOS even with sudo privileges, so it should be safe
+# to add it to PATH even when the application wants to use a very strict PATH.
```
rrbutani, line 72:
```suggestion
# On macOS, we use ld as the linker for single-platform builds (i.e. when not
# cross-compiling). Some rule sets may drop `/usr/bin` from $PATH before calling
# this script, which makes `/usr/bin/ld` unreachable.
#
# For example, `rules_rust` does this:
# https://github.com/bazelbuild/rules_rust/blob/e589105b4e8181dd1d0d8ccaa0cf3267efb06e86/cargo/cargo_build_script.bzl#L66-L68
# 
# Let's restore `/usr/bin` to $PATH in such cases. Note that `/usr/bin` is not a
# writeable directory on macOS even with sudo privileges, so it should be safe
# to add it to $PATH even when the application wants to use a very strict $PATH.
```

rrbutani:
> `rules_rust` [actually](https://github.com/bazelbuild/rules_rust/blob/52e3ba4c99692b5354faff0fb1aa62ddf884024c/rust/private/rustc.bzl#L232-L236) uses [`cc_common.get_environment_variables`](https://docs.bazel.build/versions/main/skylark/lib/cc_common.html#get_environment_variables) to get the env vars to use in its invocation; I wonder if there's a way to get that function to keep `/usr/bin` in the `$PATH`.

It's really `cc_common.create_link_variables` which calls [this](https://github.com/bazelbuild/bazel/blob/506cf1be7eb08b674023937bd6939baa0a512afe/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkBuildVariables.java#L99).

rrbutani:
> It's really `cc_common.create_link_variables` which calls [this](https://github.com/bazelbuild/bazel/blob/506cf1be7eb08b674023937bd6939baa0a512afe/src/main/java/com/google/devtools/build/lib/rules/cpp/LinkBuildVariables.java#L99).

Ultimately these come from the `env_sets` on `feature`s (i.e. grep for `env_entries` in [this file](https://github.com/bazelbuild/bazel/blob/7dd6ced654e0ed96e0047c369288f283481c7265/tools/osx/crosstool/cc_toolchain_config.bzl)) which (predictably, understandably, etc.) don't influence `$PATH`.

Still trying to figure out how/where `$PATH` actually gets set in `rules_rust`.

rrbutani:
> Still trying to figure out how/where `$PATH` actually gets set in `rules_rust`; it's not coming from `link_env`.

$PATH really is unset when `rules_rust` calls its process wrapper and it's still unset when the process wrapper calls `rustc`.

Not sure where along the way from `Bazel` -> `process_wrapper` -> `rustc` -> `cc_wrapper.sh` -> `clang` `$PATH` is getting set again but it doesn't _really_ matter. It makes sense that `$PATH` is essentially getting set to `$PWD`.

Adding `/usr/bin` back into the `$PATH` feels a little hacky but it doesn't actually pose any problems and I can't think of a better solution.

rrbutani:
Specifying `-B /usr/bin` works but I don't know if that's really any better.

siddharthab:
If `rules_rust` does not provide PATH in the env it is using for `rustc` (evident when you run `bazel build -s` and see that PATH is missing from the env in the invocation of process-wrapper, and confirm that their process-wrapper [also does not set PATH](https://github.com/bazelbuild/rules_rust/blob/f0799954981040e145415d6b78a0d7e697ea9a68/util/process_wrapper/process_wrapper.cc#L32-L37) in the env it is using for `rustc`), then [this block](https://github.com/rust-lang/rust/blob/1c03f0d0ba4fee54b7aa458f4d3ad989d8bf7b34/compiler/rustc_codegen_ssa/src/back/link.rs#L640-L645) in `rustc` will set the path to "/bin".

siddharthab:
One more thing. To confirm that the expected behavior would happen if PATH was not set to `/bin`:
```
% env -i /bin/bash -c 'echo ${PATH}'
/usr/gnu/bin:/usr/local/bin:/bin:/usr/bin:.
% env -i PATH=/bin /bin/bash -c 'echo ${PATH}'
/bin
```

toolchain/osx_cc_wrapper.sh.tpl
siddharthab: I have now added more explanation.

rrbutani:
@siddharthab When you add `echo $PATH > /tmp/whatever` to `cc_wrapper.sh` do you actually get `PATH=/bin`? I just get the Rust toolchain bin dir (i.e. `/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/bin` on my machine).

rrbutani:
> @siddharthab When you add `echo $PATH > /tmp/whatever` to `cc_wrapper.sh` do you actually get `PATH=/bin`? I just get the Rust toolchain bin dir (i.e. `/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/bin` on my machine).

^ is actually consistent with what [this code snippet](https://github.com/rust-lang/rust/blob/1c03f0d0ba4fee54b7aa458f4d3ad989d8bf7b34/compiler/rustc_session/src/session.rs#L804-L813) is doing: `sysroot` + `rustlib` + `bin`

-----

This [snippet](https://github.com/rust-lang/rust/blob/1c03f0d0ba4fee54b7aa458f4d3ad989d8bf7b34/compiler/rustc_codegen_ssa/src/back/link.rs#L640-L645) is a good find but it only _extends_ `$PATH`. Together with ^ it explains the behavior we're seeing; `$PATH` is unset until we reach `cc_wrapper.sh` where it's set (by `rustc` as you correctly surmised).

rrbutani:
> One more thing. To confirm that the expected behavior would happen if PATH was not set to `/bin`:
> 
> ```
> % env -i /bin/bash -c 'echo ${PATH}'
> /usr/gnu/bin:/usr/local/bin:/bin:/usr/bin:.
> % env -i PATH=/bin /bin/bash -c 'echo ${PATH}'
> /bin
> ```

Sorry I'm not sure if I'm understanding correctly; are you saying the `$PATH` would still contain `/usr/bin` if not for `rustc`?

As mentioned, `rustc` is extending `$PATH` and `$PATH` is unset to begin with. If you replace the process wrapper in `rules_rust` with a script that dumps `env` somewhere you can observe this.

In your example, `env -i /bin/bash -c 'echo ${PATH}'` still prints out the default `${PATH}` despite `-i` because of `bash`'s defaults (if `$PATH` is unset it populates it, I think – if you run `env -i /bin/bash --norc --noprofile -c 'echo ${PATH}'` it's empty as expected).

rrbutani:
`bash` also doesn't seem to do that when running scripts; if you do `env -i bash /tmp/record` where `/tmp/record` is:

```bash
#/usr/bin/env bash

env > /tmp/dump
```

`/tmp/dump` just has `$_`, `$PWD`, and `$SHLVL` (on my machine at least)

rrbutani:
None of this ultimately changes the fix but I do think it's not accurate to blame `rustc` for the `$PATH` not having `/usr/bin` (that's because of `rules_rust` and it seems like an entirely sensible choice). You're absolutely correct that `rustc` is adding `{sysroot}/{rust_lib}/bin` to `$PATH` though.

toolchain/osx_cc_wrapper.sh.tpl
```diff
@@ -62,6 +62,26 @@ for i in "$@"; do
     fi
 done
 
+# On macOS, we use ld as the linker for single-platform builds (i.e., when not
+# cross-compiling). Some applications may remove /usr/bin from PATH before
+# calling this script, which would make /usr/bin/ld unreachable.  For example,
+# rules_rust does not set PATH (unless the user explicitly sets PATH in env
+# through attributes) [1] when calling rustc, and rustc explicitly sets PATH to
+# be the PATH from its environment and {sysroot}/{rust_lib}/bin ([2], [3]),
```
rrbutani, line 70:
```suggestion
# through attributes) [1] when calling rustc, and rustc explicitly adds
# `{sysroot}/{rust_lib}/bin` to `$PATH` for macOS hosts ([2], [3]),
```

toolchain/osx_cc_wrapper.sh.tpl
```diff
@@ -62,6 +62,26 @@ for i in "$@"; do
...
+# which results in CC being called with PATH={sysroot}/{rust_lib}/bin. Note
+# that leaving PATH unset would make bash use the default PATH value which does
+# contain /usr/bin, so from the perspective of this bash wrapper script, PATH
```
rrbutani, line 73: Last nit, sorry.

I think this actually isn't true for scripts; i.e. running `env -i bash cc_wrapper.sh` would cause `$PATH` to be unset for this script.

siddharthab:
My apologies if it felt like I was blaming rustc. I don't think either rules_rust or rustc are doing the wrong thing. That is why I did not advise filing an issue. Ideally, users should be setting `--action_env=PATH=....` and then rules should be honoring that (I did not confirm this behavior with rules_rust).

I also don't think that symlinking is a worse solution. It is actually a more precise solution and is guaranteed to work no matter what the value of PATH is. I was merely trying to explain why only the linker in `rules_rust` was not working, whereas it was working for `rules_cc` and `rules_go`, and my main thought in pursuing this solution was that it makes life easier for people who are using the `toolchain_roots` attribute with `http_archive` or `local_repository` based LLVM distributions.

I don't have a lot of time to give this, so happy to just abandon and leave things at the current state.

I don't know how to read Rust, so on reading that code fragment, I initially thought that the PATH would be set to /bin, but I was wrong. I have changed my comment to reflect the new understanding.

> @siddharthab When you add `echo $PATH > /tmp/whatever` to `cc_wrapper.sh` do you actually get `PATH=/bin`? I just get the Rust toolchain bin dir (i.e. `/private/var/tmp/_bazel_rahulbutani/729ef14ca5fc8012394a84396ed546ff/external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/bin` on my machine).

That's also what I get (I had a misunderstanding before). Across all build actions for `@git2:all`:
```
   4 PATH=_my_env_path_
  15 PATH=..../external/rust_darwin_x86_64/lib/rustlib/x86_64-apple-darwin/bin
 219 PATH=/usr/gnu/bin:/usr/local/bin:/bin:/usr/bin:.
```

> `bash` also doesn't seem to do .... where `/tmp/record` is:
> ```shell
> #/usr/bin/env bash
> 
> env > /tmp/dump
> ```

I am not claiming that env contains PATH (it is still unset). I am saying that bash will use a default value when finding programs. Instead of `env > /tmp/dump`, try `echo "$PATH"` or even better, just run ld itself: `ld -v`.

toolchain/osx_cc_wrapper.sh.tpl
siddharthab: Explained in the other comment.

rrbutani:
> My apologies if it felt like I was blaming rustc.

oh not at all!

Using the word "blame" was a poor choice on my part. I was just trying to say that I didn't think `rustc` was resetting the `$PATH` and wasn't trying to make any kind of judgement about the "source" of the problem.

> I also don't think that symlinking is a worse solution. It is actually a more precise solution and is guaranteed to work no matter what the value of PATH is. I was merely trying to explain why only the linker in `rules_rust` was not working, whereas it was working for `rules_cc` and `rules_go`, and my main thought in pursuing this solution was that it makes life easier for people who are using the `toolchain_roots` attribute with `http_archive` or `local_repository` based LLVM distributions.

I absolutely agree! I think that reason alone is a compelling reason to make this change.

> > `bash` also doesn't seem to do .... where `/tmp/record` is:
> > ```shell
> > #/usr/bin/env bash
> > 
> > env > /tmp/dump
> > ```
> 
> I am not claiming that env contains PATH (it is still unset). I am saying that bash will use a default value when finding programs. Instead of `env > /tmp/dump`, try `echo "$PATH"` or even better, just run ld itself: `ld -v`.

oh good catch! That's very interesting behavior; I didn't realize `bash` would use a default `$PATH` when `$PATH` without it actually being visible to subprocesses (like `env`).

I think in our specific case this bit of behavior actually doesn't save us. Running `ld -v` in `/tmp/record` totally works, as you mentioned, but running `clang -v -xc++ - <<<"int main() { }"` does not – because the default `$PATH` value isn't passed down to `clang`, when `clang` goes to run `ld` it still fails.

rrbutani:
> I don't have a lot of time to give this, so happy to just abandon and leave things at the current state.

Sorry, I didn't mean to make this more difficult; I just really wanted to get to the bottom of what was causing the issue.

Regardless, I think we have a pretty good understanding now and I'm definitely happy with this PR as is.

siddharthab:
> I think in our specific case this bit of behavior actually doesn't save us. Running `ld -v` in `/tmp/record` totally works, as you mentioned, but running `clang -v -xc++ - <<<"int main() { }"` does not – because the default `$PATH` value isn't passed down to `clang`, when `clang` goes to run `ld` it still fails.

Thanks. I have now removed that comment. I think then things were working elsewhere because in all other rules, PATH is retained from the user environment or set to a reasonable default.

siddharthab:
If I run bazel with `--experimental_strict_action_env`, all actions are run with an empty environment. In this case,
- rules_rust fails because of what we know.
- rules_cc works because it [sets](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java;l=529?q=%22%2Fusr%2Flocal%2Fbin%22&ss=bazel%2Fbazel) the default path to `/bin:/usr/bin:/usr/local/bin`
- rules_go works because it [ensures](https://github.com/bazelbuild/rules_go/blob/63dfd99403076331fef0775d52a8039d502d4115/go/private/context.bzl#L434) /usr/bin is in PATH.

siddharthab:
I think this PR is now OK to be merged if @dozzman is OK with this approach. Will wait for them to comment.

dozzman (approved):
Yep this is perfect.

PR #107, opened 2021-09-27
Make platforms targets publicly visible

dududko:
Make targets in the platforms package publicly visible so they could be used in the attributes of other targets. 
Example:
```
with_platform(
    name = "linux",
    platform = "@com_grail_bazel_toolchain//platforms:linux-x86_64",
    tool = ":binary",
    visibility = ["//visibility:public"],
)
```

### DIFF
--- a/platforms/BUILD.bazel
+++ b/platforms/BUILD.bazel
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+package(default_visibility = ["//visibility:public"])
+
 platform(
     name = "linux-x86_64",
     constraint_values = [


rrbutani:
Thanks for the PR!

Can you tell me a little bit more about your use case? `//platforms` really only exists [for our own tests](https://github.com/grailbio/bazel-toolchain/blob/ddcdd61f64908b72ff9805f97863ce83ab4a21d4/tests/scripts/run_xcompile_tests.sh#L36); I'm a bit wary to have users depend on the stuff there.

In my mind, users would either just use the auto-generated `@local_config_platform//:host` as their host platform (and let toolchain resolution pick the appropriate toolchain) or define their own platform when cross-compiling or doing something a bit more special.

In your example, what does `with_platform` do?

dududko:
There are two way to do cross-compilation:
- by specifying target platform with `--platform` option
- using [transition](https://docs.bazel.build/versions/main/skylark/lib/transition.html)

There is a third option in rules_go, where you can specify goos/goarch attributes in go_binary, but this approach is actually based on `transitions`.

In my case I need to build binary for multiple platforms (darwin/linux) and publish build artifacts. I prefer doing this in a single bazel executable target without specifying `--platform` option on each build, thus I use transitions. A have a very simple rule for this which is called `with_platform`, and all this rules does is apply a target platform in the build configuration of a `tool` attribute, which can be any binary target.

I use this approach when build a custom gazelle `gazelle_binary ` for multiple platforms. This target does not support goos/goarch [anymore](https://github.com/bazelbuild/bazel-gazelle/issues/803#issuecomment-645646692), so I had to implement a custom transition. And I use the same transition for building go_binary with cgo for multiple platforms using the toolchains that you provide.

> I'm a bit wary to have users depend on the stuff there.

In [rules_go](https://github.com/bazelbuild/rules_go/blob/master/go/toolchain/BUILD.bazel#L7-L9) platforms are also publicly visible, and users depend on those targets. I do no see the reason why not to make your platform targets visible as well.

rrbutani:
> In [rules_go](https://github.com/bazelbuild/rules_go/blob/master/go/toolchain/BUILD.bazel?rgh-link-date=2021-09-27T15%3A51%3A45Z#L7-L9) platforms are also publicly visible, and users depend on those targets. I do no see the reason why not to make your platform targets visible as well.

My hesitation mostly comes from the fact that while we have a small list of supported target/host platforms at the moment, we eventually hope to support arbitrary target triples in which case we'd want to generate the `platform`s (as #85 does). Under that scenario it's potentially problematic to have users depend on the generated platforms since they'd be tied to the value of `extra_targets` that's provided (I'd hope that, for example, `@llvm_toolchain//platforms:wasi` only being present if `wasm32-unknown-wasi` is specified as an extra target is expected and reasonable behavior but I'm not sure this is the case; I think it'd be easy to accidentally assume the generated platforms are always present and to use them for things that have nothing to do with this toolchain).

That said, I see the utility in providing these and I think it makes sense to expose these platforms for now with the caveat that they may move to the generated toolchain repo in a future release (0.7+).

platforms/BUILD.bazel
```diff
@@ -18,6 +18,7 @@ platform(
         "@platforms//os:linux",
         "@platforms//cpu:x86_64",
     ],
+    visibility = ["//visibility:public"],
```
rrbutani, line 21: Small nit: can we add a `package(default_visibility = ["//visibility:public"])` to the top of this file and remove the individual `visibility` attributes?

dududko: changed

PR #112, opened 2021-10-11
Allow selection of stdlib to use

jsharpe:
Adds a `use_libcpp` attribute to the `llvm_toolchain` repository rule to allow selection of whether `libc++` or `stdc++` is used.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -51,7 +51,9 @@ filegroup(
 llvm_toolchain(
     name = "llvm_toolchain_with_sysroot",
     llvm_version = "12.0.0",
-    stdlib = "stdc++",
+    stdlib = {
+        "linux-x86_64": "stdc++",
+    },
     sysroot = {
         "linux-x86_64": "@org_chromium_sysroot_linux_x64//:sysroot",
     },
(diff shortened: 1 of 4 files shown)


siddharthab:
Hello, thank you for the PR. I just have one thought. Should we take in a string attribute called `stdlib` that could be one of `bundled-libc++` (default), `libc++`, or `libstdc++`. This way, there can be more options in the future. For example, when cross-compiling, how to get the one for the target architecture.

jsharpe:
I've updated this PR to follow what was discussed in the review comments but I haven't got time to look into how to resolve the failure in the linux_sysroot case. Any ideas what needs to be done to fix up that case?

siddharthab:
> linux_sysroot case. Any ideas what needs to be done to fix up that case?

I have added a temporary measure to always include libpthread and libdl, to keep parity between the different stdlib options. But I think we should make people explicitly link in libpthread if they need it.

siddharthab (approved):

rrbutani (approved):

siddharthab:
I am going to land this now as we discussed in this thread. But I may revert some of the design here when I work on the long term configurability today. I am thinking of directly exposing the compile_flags and link_flags through the repository rule. This will give people the ability to change things and even override some of the defaults.

siddharthab:
I have exposed overwriting the compile_flags and link_flags through repository_rule attributes but decided to keep the changes as implemented for this issue, because it is a common enough feature request and overriding all the flags for this is not as smooth as I would like it to be.

PR #116, opened 2021-10-15
Use exec to call the compiler

jsharpe:


rrbutani:
Oh, of course; my bad 🤦

PR #123, opened 2021-11-12
Always use ubuntu16.04 binaries for arch linux

DCNick3:
Using ubuntu20.04 binaries on arch linux has an unfortunate side-effect of printing `external/llvm_toolchain_llvm/bin/clang: /usr/lib/libtinfo.so.6: no version information available (required by external/llvm_toolchain_llvm/bin/clang)` for each invocation of clang. While these warnings do not pose any danger except being an annoyance, there is no reasonably simple way to suppress them (see https://refi64.com/posts/qldv.html). Binaries for ubuntu16.04 do not have this problem though and seem to work fine, although requiring an `ncurses5-compat-libs` package

I am not sure if this change is appropriate for the master, as it requires additional dependencies to be installed. Lower versions of LLVM would require this package too, though

### DIFF
--- a/tests/scripts/archlinux_test.sh
+++ b/tests/scripts/archlinux_test.sh
@@ -34,7 +34,7 @@ Server = https://repo.archlinuxcn.org/\$arch
 EOF
 
 # Install dependencies
-pacman -Syu --noconfirm --quiet python ncurses5-compat-libs
+pacman -Syu --noconfirm --quiet archlinuxcn-keyring python ncurses5-compat-libs
 
 # Run tests
 cd /src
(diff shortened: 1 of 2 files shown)


DCNick3:
#53 probably should be taken into consideration..

siddharthab:
I think I am going to allow users to specify an OS mapping of their own in the repository rule. That way, we won't have to keep going back and forth in the script.

I think the warnings should eventually be handled within llvm itself. We can not rely on continued distributions meant for an OS (Ubuntu 16.04) past its EOL.

I am going to close this out now, but please feel free to reopen if you have more thoughts.

PR #129, opened 2021-12-18
Allows customized c++ standard.

yaxum62:
Allows user specify c++ standard using `cxx_standard` attribute of `llvm_toolchain` rule, instead of hardcoded "c++17".


siddharthab:
I am going to merge this change as-is now. But will probably revert this later today when I expose compile_flags and link_flags more directly so people can override these settings. Need to think it out a little more.

siddharthab:
I have exposed overwriting the compile_flags through a repository_rule attribute but decided to keep the changes as implemented for this issue, because it is a common enough feature request and overriding all the flags for this is not as smooth as I would like it to be.

PR #136, opened 2022-02-18
Bazel: Added support for openmp

storypku:
Otherwise, bazel build //tests:omp_test will fail:

```
ld.lld: error: unable to find library -lomp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Target //tests:hello_omp failed to build
```

However, it still requires "libomp.so" can be found in ldconfig path during runtime.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -31,6 +31,13 @@ cc_test(
     deps = [":stdlib"],
 )
 
+cc_test(
+    name = "omp_test",
+    srcs = ["omp_test.c"],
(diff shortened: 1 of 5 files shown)


siddharthab:
Thank you for the PR.

I think it is not so straightforward here on what is the best path with dynamic libraries.

Since we can not static link libomp and will require a runtime installation, it is better to use libomp through `deps`. At GRAIL, we use a `cc_import` target to use the bundled libomp. I can provide a convenience target in this project, so people can use directly.

PR #142, opened 2022-04-06
fix: missing include and lib search paths

f0rmiga:
Added some missing search paths. Getting errors to build some software without them.

### DIFF
--- a/toolchain/BUILD.llvm_repo
+++ b/toolchain/BUILD.llvm_repo
@@ -41,7 +41,7 @@ filegroup(
 filegroup(
     name = "include",
     srcs = glob([
-        "include/c++/**",
+        "include/**/c++/**",
         "lib/clang/*/include/**",
     ]),
 )
@@ -55,7 +55,7 @@ filegroup(
     name = "lib",
     srcs = glob(
         [
-            "lib/lib*.a",
+            "lib/**/lib*.a",
             "lib/clang/*/lib/**/*.a",
         ],
         exclude = [
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -174,6 +174,7 @@ def cc_toolchain_config(
             # libraries.
             link_flags.extend([
                 "-L{}lib".format(toolchain_path_prefix),
+                "-L{}lib/{}".format(toolchain_path_prefix, target_system_name),
                 "-l:libc++.a",
                 "-l:libc++abi.a",
                 "-l:libunwind.a",
@@ -232,6 +233,7 @@ def cc_toolchain_config(
     # C++ built-in include directories:
     cxx_builtin_include_directories = [
         toolchain_path_prefix + "include/c++/v1",
+        toolchain_path_prefix + "include/{}/c++/v1".format(target_system_name),
         toolchain_path_prefix + "lib/clang/{}/include".format(llvm_version),
         toolchain_path_prefix + "lib64/clang/{}/include".format(llvm_version),
     ]


rrbutani:
@f0rmiga Can you provide some more detail? What library/include paths are missing for you?

In [official release tarballs](https://github.com/llvm/llvm-project/releases), the only `c++` directory I can find in `include` with headers is `include/c++`; ditto with `lib` and `.a` files (just `lib` and `lib/clang/<ver>/lib/...`).

Based on:
https://github.com/grailbio/bazel-toolchain/blob/5fe1b6694b544e7cfcf7550f83f2707d25fff053/toolchain/cc_toolchain_config.bzl#L177
it sounds like you're maybe cross-compiling + looking to use libraries on your local system?

f0rmiga:
The official 14.0.0 release for Linux x86_64:

```bash
clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04 $ find . -type d -name '*c++*'
./include/c++
./include/x86_64-unknown-linux-gnu/c++

clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04 $ find . -type f -name 'libc++*.a'
./lib/x86_64-unknown-linux-gnu/libc++.a
./lib/x86_64-unknown-linux-gnu/libc++experimental.a
./lib/x86_64-unknown-linux-gnu/libc++abi.a
```

f0rmiga:
I wonder how https://github.com/grailbio/bazel-toolchain/blob/32d7596dd99c608015ba9a20c24b4fd5d03fdad2/toolchain/cc_toolchain_config.bzl#L177-L179 would find these libraries without `"-L{}lib/{}".format(toolchain_path_prefix, target_system_name)`.

rrbutani:
Looking at some release tarballs, this seems like a change that was made in 14.0.0 but only for _some_ platforms (so far only SLES12 and x86_64 Ubuntu 18.04 seem to have this subdir):

```
❯ fd 'libc\+\+.a'
clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/lib/libc++.a
clang+llvm-14.0.0-amd64-unknown-freebsd13/lib/libc++.a
clang+llvm-14.0.0-armv7a-linux-gnueabihf/lib/libc++.a
clang+llvm-14.0.0-powerpc64le-linux-ubuntu-18.04/lib/libc++.a
clang+llvm-14.0.0-x86_64-apple-darwin/lib/libc++.a
clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/lib/x86_64-unknown-linux-gnu/libc++.a
clang+llvm-14.0.0-x86_64-linux-sles12.4/lib/x86_64-unknown-linux-gnu/libc++.a
```

I can't tell if this was an intentional change or not; I haven't yet looked into how the release tarballs are actually put together.

-----

Regardless, if this is how some of the 14.0.0 tarballs are laid out we should change `bazel-toolchain` to match.

 My concern is that (IIRC) `ld` warns if library directories it's given don't exist. I think this is worth avoiding (flooding users that aren't on 14.0.0 + particular platforms with warnings seems pretty suboptimal). Ideally we'd gate adding that `-L` flag on the existence of `lib/<target-triple>` but (afaik) there's no clean way to do this from within `cc_toolchain_config`. We'd need to do the check in a repository rule (`llvm_config`) and thread through the information.

The alternative is to hardcode the list of version+platforms that have a target subdirectory but I think that's a last resort.

siddharthab:
I have not given too much thought to this, but one suggestion could be to also allow a `{target_system_name}` placeholder in the cxx and link flags attributes at:
https://github.com/grailbio/bazel-toolchain/blob/32d7596dd99c608015ba9a20c24b4fd5d03fdad2/toolchain/rules.bzl#L163

This way, any user who finds themselves in this situation will be able to easily customize the link_flags for their use.

As for good defaults that work for all LLVM release distributions, I don't think there is any harm in adding these locations for non-macOS distributions. AFAIK, lld does not complain about non-existent directories by default, but ld64 on macOS does. But I could be wrong. If lld is complaining about this, then maybe just the above solution of having the users set the link_flags themselves, or setting it up ourselves based on the distribution name (and documenting it in the README) could be OK.

f0rmiga:
For the GCC toolchain I wrote [here](https://github.com/aspect-build/gcc-toolchain/blob/c7bc6e55dbf5bd246562da81f5f4dc627165f396/toolchain/config.bzl.tpl), I made it a template that I render from the repository rule. This is one way of passing the information we only have access in the repository rule to the config.

siddharthab:
We used to do that before, but it led to very expensive repository refetches (mostly extraction of the downloaded and cached archive file) of the toolchain archive when some things changed. So we separated the configuration and the distribution repos.

djmarcin:
I encountered this today as well. I made the same fix locally but only to the `BUILD.llvm_repo` file. For whatever reason I did not have to update the library search paths in `cc_toolchain_config.bzl`. Was that a necessary update for you @f0rmiga? I'm wondering if LLVM automatically knows to look in these platform specific dirs somehow. I'm only building Rust so I can imagine it's needed for C++ perhaps.

Are there more things that need to be solved before this can be merged?

rrbutani:
> I encountered this today as well.

@djmarcin are you also using LLVM 14?

> I made the same fix locally but only to the `BUILD.llvm_repo` file. For whatever reason I did not have to update the library search paths in `cc_toolchain_config.bzl`. Was that a necessary update for you @f0rmiga? I'm wondering if LLVM automatically knows to look in these platform specific dirs somehow. I'm only building Rust so I can imagine it's needed for C++ perhaps.

This is interesting, I'm not really sure what's going on here. Before you added the target specific dirs to the filegroups in `BUILD.llvm_repo` were you getting linker errors about `libc++` being missing?

Just to be sure, are you using [`builtin-libc++`](https://github.com/grailbio/bazel-toolchain/blob/32d7596dd99c608015ba9a20c24b4fd5d03fdad2/toolchain/rules.bzl#L125-L134)?

> Are there more things that need to be solved before this can be merged?

As mentioned [here](https://github.com/grailbio/bazel-toolchain/pull/142#issuecomment-1091824723) I want to try checking for the existence of the target specific dir in the repository rule; if that proves too tricky we can fall back to [the alternatives @siddharthab described](https://github.com/grailbio/bazel-toolchain/pull/142#issuecomment-1092198501).

I'll put something together later today.

siddharthab:
Sent 869212179aff3176e48ba6ee891ce58cead100d3 to update the tested version to be 14. We will need to fix the tests in subsequent revisions.

siddharthab:
Verified that we don't need the -L flags at all. clang inserts them if the files are present in the sandbox. Same thing for the include dir; if the files are in the sandbox, clang picks up the dir.

Verified using (this PR rebased on current master, and on my Ubuntu 20.04 machine):
```
cd tests
bazel build -s --copt=-v --linkopt=--verbose --linkopt=-Wl,--verbose //:stdlib
```

siddharthab:
I am going to merge this change and then remove the -L flags and the builtin_include_dirs in a separate commit.

djmarcin:
@rrbutani It looks like this is sorted out but just to close the loop on some questions -- yes I am using LLVM14 and I was getting these errors:
```
  = note: ld.lld: error: unable to find library -l:libc++.a
          ld.lld: error: unable to find library -l:libc++abi.a
          ld.lld: error: unable to find library -l:libunwind.a
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

siddharthab:
Closing the loop. All tests have now been fixed to support LLVM 14, except SUSE (see #143).

Issue #145, opened 2022-05-18
v0.7.1 breaks Bazel 4.2.2


zhanyong-wan:
I tried to set up bazel-toolchain 0.7.1 with Bazel 4.2.2, and got an error

```
ERROR: .../external/com_grail_bazel_toolchain/toolchain/internal/configure.bzl:391:5: nested functions are not allowed. Move the function to the top level.
```

I'm using Bazel 4.2.2:

```
$ bazel version
Build label: 4.2.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 2 18:15:58 2021 (1638468958)
Build timestamp: 1638468958
Build timestamp as int: 1638468958
```

README.md says that the minimum Bazel version supported is 4.0.0, but it seems that 4.2.2 doesn't understand nested functions yet.

Could https://github.com/grailbio/bazel-toolchain/commit/c0e0937e63b8adbd04f89782e171103a1590e33b be the culprit?

zhanyong-wan:
Sent the PR: https://github.com/grailbio/bazel-toolchain/pull/146

siddharthab:
Thanks!

zhanyong-wan:
Thanks for the quick review, [Siddhartha](https://github.com/siddharthab)!  Do you plan to cut a new release soon to include this fix?  Thanks!

siddharthab:
Working on the release today. I want to fix the archlinux container test first.

siddharthab:
A new tag is now out. You have also been invited to have Write permissions on this repo.

zhanyong-wan:
Thank you!!

PR #146, opened 2022-05-20
Avoid nested skylark functions as they don't work with bazel 4.*

zhanyong-wan:


zhanyong-wan:
This PR should fix https://github.com/grailbio/bazel-toolchain/issues/145 - please review.  Thanks!

PR #148, opened 2022-06-02
Treat Pop!_OS as ubuntu

zaucy:
pop os suppor was broken for me as I am currently running `22.04`

#62 seemed to mix pop os with linuxmint and specific version checks, but AFAIK pop os uses the same version as ubuntu. This change simply treats pop as ubuntu.


siddharthab:
Thanks!

PR #159, opened 2022-11-06
Expose the `target_settings` attribute for toolchains

philsc:
This is useful when letting users select between multiple toolchains.


siddharthab:
Thanks. I did not know about this attribute.

PR #163, opened 2022-11-25
Update BUILD.llvm_repo for building static linked binary

BoleynSu:


siddharthab:
Hi! Thank you for your contribution. Can you give me an example of when you would need .o files instead of the .a files?

BoleynSu:
I cannot really remember it now. IIRC, it is related to build statically linked binary with ptheead. I can try repro it if you insist.

siddharthab:
Yes please if you want this change merged. The use case is not obvious to me and since you can not recollect it either, it may not be very straightforward. So it would be good to add a unit test or at least some documentation to protect this change from future refactoring.

BoleynSu:
You can repro it by deleting the .o files shipped by the system. Let me update the fedora test alone with the change.

tests/scripts/fedora_test.sh
```diff
@@ -24,12 +24,11 @@ readonly git_root
 
 for image in "${images[@]}"; do
   docker pull "${image}"
-  docker run --rm --entrypoint=/bin/bash --volume="${git_root}:/src:ro" "${image}" -c """
+  docker run --rm --entrypoint=/bin/bash --volume="${git_root}:/src:ro,z" "${image}" -c """
```
siddharthab, line 27: Why `-z` here?

tests/scripts/run_tests.sh
```diff
@@ -39,6 +39,7 @@ test_args=(
   --extra_toolchains="${toolchain_name}"
   --copt=-v
   --linkopt=-Wl,-t
+  --crosstool_top=//:fake_crosstool_top
```
siddharthab, line 42: Why the need for a fake crosstool?

siddharthab: I think this can be achieved with exporting BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 in the environment.

toolchain/BUILD.llvm_repo
```diff
@@ -57,6 +57,7 @@ filegroup(
         [
             "lib/**/lib*.a",
             "lib/clang/*/lib/**/*.a",
+            "lib/clang/*/lib/**/*.o",
```
siddharthab, line 60: I see it's picking up the C runtime object files from the GCC install if we don't provide these in the sandbox. Thanks for reporting these. I mistakenly assumed that they would also be part of some .a archive.

siddharthab:
Thanks for this PR. I think this is now completely handled by commit 2c14d3ffdf9e3403f27f3b3d254734aecdea1c34. Please open a new PR if I missed something.

tests/scripts/fedora_test.sh
BoleynSu: To make it work with SELinux

tests/scripts/run_tests.sh
BoleynSu: Some other tool depends on this lagacy feature. Did not realize that this test do not. Good catch!

PR #170, opened 2023-01-06
Support LLVM 15.0.0 and 15.0.6.

zhanyong-wan:
Support LLVM 15.0.0 and 15.0.6. Also fix a failure in utils/llvm_chec…ksums.sh on Ubuntu.

### DIFF
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -15,9 +15,12 @@
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc", "use_netrc")
 load("//toolchain/internal:common.bzl", _attr_dict = "attr_dict", _python = "python")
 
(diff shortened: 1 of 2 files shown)


zhanyong-wan:
The test failure doesn't seem related to my PR.  Is it a known issue?

siddharthab:
The test failures seem to be separate. I will look into them later. They are meant to fail when we are not compatible with changes in external dependencies.

For your note on compression support in wget, I have removed it. I may have been experimenting with something when I added those lines, but because the download archive is already compressed, that flag should not have much of an effect. There were some other leftovers in that commit from my experimentation. Thanks for noticing it and taking the time to say something.

siddharthab:
The failing tests are now fixed on master through 6e282e9b760520b238a1564989141106a1b7b794.

siddharthab:
0.8.1 is now out and has this change.

PR #171, opened 2023-01-06
Define aliased tools as binaries

oliverlee:
Define tools as sh_binary's instead of aliases to filegroups, allowing use with rules that expect `files_to_run.executable` attributes (e.g. https://github.com/erenon/bazel_clang_tidy).

### DIFF
--- a/toolchain/BUILD.toolchain.tpl
+++ b/toolchain/BUILD.toolchain.tpl
@@ -14,6 +14,7 @@
 
 package(default_visibility = ["//visibility:public"])
 
+load("@bazel_skylib//rules:native_binary.bzl", "native_binary")
 load("@rules_cc//cc:defs.bzl", "cc_toolchain", "cc_toolchain_suite")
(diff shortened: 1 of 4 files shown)


oliverlee:
@rrbutani It looks like the failures in CI are unrelated to this change. Are these known to be flaky?

rrbutani:
@oliverlee Yup, the external tests are flaky. There were [some related changes recently](https://github.com/oliverlee/bazel-toolchain/compare/tool-binaries...grailbio:bazel-toolchain:master) that I think your branch is missing but even with those changes the external tests will [still fail](https://github.com/grailbio/bazel-toolchain/actions/runs/3853563617/jobs/6566658007).

I wouldn't worry about it and this PR looks good to me but @siddharthab should take a look too (I'm not really an active maintainer of this repo anymore).

siddharthab:
Can you rebase your branch on the latest master? I think at least one of those failure is gone by now.

But independent of the test failure, I would not wrap the aliased file in a sh_binary. If one does not exist, maybe making a small custom rule would do the job.

PR #172, opened 2023-01-06
Add sanitizer features

oliverlee:


siddharthab:
We will rely on https://github.com/bazelbuild/bazel/pull/17083 for this.

oliverlee:
Sounds good.

PR #176, opened 2023-01-16
fix _ubuntu_osname for 15.0.6

garymm:


garymm:
@siddharthab do you know why the macos tests are failing? I've seen something similar locally but I don't understand the root cause at all. Should I open an issue about it?

siddharthab:
I think it's because the osx wrapper script is out of sync with the toolchain config in bazel 6.0.0. I will fix this today.

PR #181, opened 2023-01-18
Remove python dependency

fionera:
Everything thats done with the python scripts is possible by using plain starlark and one call to `cat`.


toolchain/internal/release_name.bzl
```diff
@@ -138,17 +123,18 @@ def _linux(llvm_version, distname, version, arch):
         arch = "armv7a"
         os_name = "linux-gnueabihf"
     elif distname == "rhel":
-        if 8 <= float(version) < 9:
+        if 8 <= float(version) and float(version) < 9:
```
fionera, line 126: Did I miss a change while rebasing?

siddharthab:
Thanks! This is cool.

PR #187, opened 2023-02-07
fix arch in darwin release name

garymm:
Without this using on arm64 macOS fails with:
`Error in fail: Unknown LLVM release: clang+llvm-15.0.6-aarch64-apple-darwin21.0.tar.xz`


garymm:
@siddharthab PTAL

siddharthab:
I think I tested it with the previous code and explicitly paid attention to this, so was a little surprised by this change. But maybe I am misremembering things. Let's commit this for now and I will test again when I have time.

PR #188, opened 2023-02-11
Add all_includes target for building loadable clang-tidy plugins

molar:
Hello

I would like to add the following filegroup "all_includes".

by exposing all includes it is possible for users of this toolchain to build clang-tidy plugins compatible with the toolchain`s clang-tidy binary, using the `-load=` option of clang-tidy.

thank you for your time.

### DIFF
--- a/toolchain/BUILD.llvm_repo
+++ b/toolchain/BUILD.llvm_repo
@@ -48,6 +48,11 @@ filegroup(
 )
 
 filegroup(
+  name = "all_includes",
+  srcs = glob(["include/**"]),
+)
+
+filegroup(
     name = "bin",
     srcs = glob(["bin/**"]),
 )


toolchain/BUILD.llvm_repo
```diff
@@ -47,6 +47,11 @@ filegroup(
     ]),
 )
 
+filegroup(
```
jsharpe, line 50: Wouldn't a `cc_library` target be a better option here?

siddharthab: This change has been inactive for a while, but I don't mind bringing it in. For `cc_library` vs `filegroup`, not sure how the end users wanted to use it, but `filegroup` seems the most flexible, so I think it's OK. I am going to merge this change as-is. Please let me know if there is a concern and I can revert.

PR #191, opened 2023-03-10
Fix call to `int(version)`.

grebe:
In 08d1aa8cee67a7a9c51c09225cead1dad42dadc3, Python scripts were rewritten in Starlark. The Python code

```
try:
    int_version = int(version)
except ValueError:
    pass
```

became

```
int_version = int(version)
```

This will fail if `version` is not an integer. Fix this by adding a check of `version.isdigit()` before calling `int()`.


siddharthab:
@jcharum I lost permissions on this repo when I left the org. Will you be able to give me write permissions as a collaborator on this and other bazel repos?

jsharpe (approved):

PR #196, opened 2023-04-11
Allow rule `llvm_toolchain` to use attribute `llvm_versions` to refer to

helly25:
Allow `llvm_toolchain` to use attribute `llvm_verions` to provide a target/version mapping. This
allows to refer to the distributions configured in `llvm_distributions.bzl`.

This way users can provide the exact mapping they require for all
targets they want to support. For instance:

llvm_toolchain(
    name = "llvm_toolchain",
    llvm_versions = {
        "": "15.0.6",
        "darwin-aarch64": "15.0.7",
        "darwin-x86_64": "15.0.7",
    },
)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,12 +70,12 @@ attributes to `llvm_toolchain`.
 
 ## Advanced Usage
 
-#### Per target LLVM version
+#### Per host architecture LLVM version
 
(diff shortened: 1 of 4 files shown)


siddharthab:
Thanks. This makes sense. I will make some cosmetic edits to the documentation and merge soon.

PR #198, opened 2023-05-15
Add support for bzlmod

gferon:
This is a fairly straightforward change that adds support for using `bazel-toolchain` with `bzlmod`. We've been using this patch on a rather large checkout with success.

The `llvm.toolchain` extension is only a wrapper around the existing `llvm_toolchain` repository rule and supports the same attributes.

For trying it out, add the following to your `MODULE.bazel`:

```
bazel_dep(name = "com_grail_bazel_toolchain", version = "0.8")
git_override(module_name = "com_grail_bazel_toolchain", remote = "https://github.com/grailbio/bazel-toolchain", commit = "cf915e5c7cfcd19a3e71de54e385e01240b865dc")

llvm = use_extension("@com_grail_bazel_toolchain//toolchain:extensions.bzl", "llvm")
llvm.toolchain(
    llvm_version = "15.0.0"
)

use_repo(llvm, "llvm_toolchain")

register_toolchains("@llvm_toolchain//:all")
```


MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(
+    name = "com_grail_bazel_toolchain",
+    version = "0.8",
```
siddharthab, line 3: Should this be version 0.8.2?

README.md
```diff
@@ -22,7 +22,24 @@ implementation, please let me know and I can redirect people there.
 
 Minimum bazel version: **4.2.1**
 
-To use this toolchain, include this section in your WORKSPACE:
+If you're using `bzlmod`, add the following to `MODULE.bazel`:
+
+```starlark
+bazel_dep(name = "com_grail_bazel_toolchain", version = "0.9")
```
siddharthab, line 28: Should this be version 0.8.2?

siddharthab (approved):
Thank you so much for this. Unsure about how version strings work in this, but otherwise all good.

MODULE.bazel
jsharpe: I would suggest setting this to 0.0.0 and using the bcr publish app to publish to the central registry which will then set this based on the tag so that commits aren't required on releases.

jsharpe: Also this should have at least 3 components as the versions are psuedo - semver.

README.md
jsharpe: Correct - it also requires this module to be published to the BCR..

MODULE.bazel
gferon: Thanks, just changed it.

toolchain/extensions.bzl
```diff
@@ -0,0 +1,28 @@
+"Module extensions for use with bzlmod"
+
+load("@com_grail_bazel_toolchain//toolchain:rules.bzl", "compiler_configuration_attrs", "llvm_repo_attrs", "llvm_toolchain")
+
+def _llvm_impl_(module_ctx):
+    for mod in module_ctx.modules:
+        for toolchain_attr in mod.tags.toolchain:
+            llvm_toolchain(
+                name = "llvm_toolchain",
```
jsharpe, line 9: This is reusing the same repo name for all tags. This will fail if multiple tags are registered.

README.md
gferon: Fixed.

toolchain/extensions.bzl
gferon: Fair point, I'll expose a better way to do this.

MODULE.bazel
```diff
@@ -0,0 +1,10 @@
+module(
+    name = "com_grail_bazel_toolchain",
```
gferon, line 2: @jsharpe / @siddharthab do you want the name of the module on the Bazel Central Registry to be `com_grail_bazel_toolchain`?

toolchain/rules.bzl
```diff
@@ -31,8 +31,8 @@ _target_pairs = ", ".join(_supported_os_arch_keys())
 _common_attrs = {
 }
 
-_llvm_repo_attrs = dict(_common_attrs)
-_llvm_repo_attrs.update({
+llvm_repo_attrs = dict(_common_attrs)
+llvm_repo_attrs.update({
```
jsharpe, line 35: nit: these probably should be moved to an internal/*.bzl file as I don't think these should be exported as public symbols from the main rules.bzl file?

toolchain/rules.bzl
```diff
@@ -97,8 +97,7 @@ _llvm_repo_attrs.update({
         doc = "An optional dict mapping host names to custom authorization patterns.",
     ),
 })
-
-_compiler_configuration_attrs = {
+compiler_configuration_attrs = {
```
jsharpe, line 100: ditto

MODULE.bazel
jsharpe: I think this is reasonable as it means that you don't have to specify `repo_name` on the `bazel_dep` to interact with other repos that already expect this name.

toolchain/rules.bzl
gferon: Fixed.

toolchain/extensions.bzl
gferon: Fixed, with default to `llvm_toolchain`.

siddharthab:
Thank you so much @jsharpe for reviewing this. I have not been active with Bazel for a few years now and am not caught up with modules. I will leave it to you to merge this.

toolchain/extensions.bzl
```diff
@@ -0,0 +1,41 @@
...
+    "@com_grail_bazel_toolchain//toolchain/internal:repo.bzl",
+    _llvm_repo_attrs = "llvm_repo_attrs",
+    _llvm_config_attrs = "llvm_config_attrs",
+)
+
+def _llvm_impl_(module_ctx):
+    for mod in module_ctx.modules:
+        for toolchain_attr in mod.tags.toolchain:
+            llvm_toolchain(
+                name = toolchain_attr.name,
```
jsharpe, line 14: Consensus is that the best practise is that for this pattern that the tag classes are restricted to being defined only from the root module to avoid conflicts in naming. See the bazel-contrib template repo for an example of how to do this.

gferon: TIL there's a template repo, amazing, thanks

gferon:
@jsharpe thanks for the thorough review, is there a way you can approve GitHub workflows for this PR so that I can iterate on the tests until everything passes?

jsharpe:
> @jsharpe thanks for the thorough review, is there a way you can approve GitHub workflows for this PR so that I can iterate on the tests until everything passes?

I don't think there is unfortunately. I think its a limitation as this is your first PR against this repo. We could get this merged and use a subsquent PR to setup CI and then I don't think we'd have the same limitation?

gferon:
> We could get this merged and use a subsquent PR

Sounds good, I know there are more adjustments to be made, for example all flags aren't forwarded to `llvm_toolchain` and CI will need adjusting. I'm definitely up for the task of finishing the work!

jsharpe:
FYI, When i've had similar failures to those in CI its been due to redefinitions of the platforms repo in the WORKSPACE.bzlmod file

fmeum:
Let me know if you want me to look into particular failures.

https://github.com/grailbio/bazel-toolchain/actions/runs/5361378105/jobs/9727258799?pr=198 is difficult to migrate as is: The `abseil-cpp` BCR module only defines the deps needed to use it, not to test it. This may require adding it as an `http_archive` in the same extension as `com_google_googletest` instead of using the BCR module.

jsharpe:
> Let me know if you want me to look into particular failures.
> 
> https://github.com/grailbio/bazel-toolchain/actions/runs/5361378105/jobs/9727258799?pr=198 is difficult to migrate as is: The `abseil-cpp` BCR module only defines the deps needed to use it, not to test it. This may require adding it as an `http_archive` in the same extension as `com_google_googletest` instead of using the BCR module.

@fmeum the errors left now are mainly on macos - I suspect the rules_go ones are due to the call to the script to generate some files before bazel runs; they're likely being placed in the wrong place when bzlmod is enabled?

Also not sure why abseil tests are failing on macOS - any insights appreciated!

gferon:
Thanks so much @jsharpe for getting this into a ready state!

MODULE.bazel
fmeum, line 2: You could use a more friendly name for this module (but don't have to).

MODULE.bazel
```diff
@@ -0,0 +1,10 @@
...
+    version = "0.0.0",
+    compatibility_level = 0,
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "rules_cc", version = "0.0.6")
+bazel_dep(name = "rules_foreign_cc", version = "0.9.0")
```
fmeum, line 9: Is this a prod dependency? If it is only used for tests, you can mark with `dev_dependency = True`.

fmeum, line 9: Is this a prod dependency? If it is only used for tests, you can mark with `dev_dependency = True`.

tests/scripts/run_external_tests.sh
```diff
@@ -27,6 +27,7 @@ cd "${scripts_dir}"
 
 test_args=(
   "${common_test_args[@]}"
+  "--experimental_enable_bzlmod=${USE_BZLMOD:-false}"
```
fmeum, line 30:
```suggestion
  "--enable_bzlmod=${USE_BZLMOD:-false}"
```
with Bazel 6 and higher.

tests/scripts/run_tests.sh
```diff
@@ -36,10 +36,12 @@ cd "${scripts_dir}"
 
 set -x
 test_args=(
-  --extra_toolchains="${toolchain_name}"
-  --copt=-v
-  --linkopt=-Wl,-t
+  "--experimental_enable_bzlmod=${USE_BZLMOD:-false}"
```
fmeum, line 39:
```suggestion
  "--enable_bzlmod=${USE_BZLMOD:-false}"
```

toolchain/extensions.bzl
```diff
@@ -0,0 +1,41 @@
+"Module extensions for use with bzlmod"
```
fmeum, line 1: It is recommended to have one module extension per file, so maybe move this to `toolchains/extensions/llvm.bzl`?

toolchain/extensions.bzl
```diff
@@ -0,0 +1,41 @@
...
+load("@com_grail_bazel_toolchain//toolchain:rules.bzl", "llvm_toolchain")
+load(
+    "@com_grail_bazel_toolchain//toolchain/internal:repo.bzl",
+    _llvm_config_attrs = "llvm_config_attrs",
+    _llvm_repo_attrs = "llvm_repo_attrs",
+)
+
+def _llvm_impl_(module_ctx):
+    for mod in module_ctx.modules:
+        for toolchain_attr in mod.tags.toolchain:
```
fmeum, line 12: I would suggest going with something like this:

```suggestion
        for toolchain_attr in mod.tags.toolchain:
            if not mod.is_root:
                fail("Only the root module can use the 'llvm' extension")
```

This avoid having to deal with the difficult questions around reconciling toolchain requirements between different modules until we have had some time to test drive this.

MODULE.bazel
jsharpe: Good catch I don't think this is required in the root MODULE.bazel file at all..

tests/scripts/run_external_tests.sh
jsharpe: 5.1.0 is still being tested in CI (but only with bzlmod disabled) so this is why this is still using the --experimental version of this flag.

MODULE.bazel
jsharpe: Hmm, thinking about it I guess its a good idea to have llvm in this name so its discoverable in the registry? I guess we could grab `llvm-toolchain` as `llvm_toolchain` is commonly used as one of the repos generated from this repo and that would avoid users having to remap repo names?

fmeum: +1 for having `llvm` in the name. Maybe still prefix with `grail` to prevent confusion if other LLVM toolchains get added (I know @illicitonion is working on one).

gferon: I'd pick `llvm_toolchain` since what we're doing is exactly and only that.

toolchain/rules.bzl
```diff
@@ -296,8 +46,17 @@ def llvm_toolchain(name, **kwargs):
             for k, v in kwargs.items()
             if (k not in _llvm_config_attrs.keys()) or (k in _common_attrs.keys())
         }
+        llvm_repository_name = name + "_llvm"
         llvm(name = name + "_llvm", **llvm_args)
-        kwargs.update(toolchain_roots = {"": "@%s_llvm//" % name})
+        if bzlmod:
+            if not bzlmod_module_version:
+                bzlmod_module_version = "override"
+            kwargs.update(toolchain_roots = {"": "@@com_grail_bazel_toolchain~{module_version}~llvm~{repository_name}//".format(
```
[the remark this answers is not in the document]
gferon: Awesome trick, I'll use that! Thanks

gferon:
I realized that this conflicts with `rules_swift` on Linux and I'm trying to understand why the toolchain is being selected instead of the `clang` from `swiftc`.

.github/workflows/tests.yml
```diff
@@ -13,17 +13,48 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest]
+<<<<<<< HEAD
```
fmeum, line 16: This is a committed merge conflict.

fmeum (approved):

.github/workflows/tests.yml
gferon: Oops, fixed!

jsharpe:
Something isn't quite right with the external_test as it seems to be running the non-bzlmod version for the bzlmod testcase.

Otherwise the changes look good to me - we're soo close to getting this landed.

tests/openssl/openssl.bazel
```diff
@@ -50,10 +50,10 @@ CONF_FILES = [
 
 genrule(
     name = "configure",
-    srcs = [("@com_grail_bazel_toolchain//tests/openssl:" + f) for f in CONF_FILES],
+    srcs = [("@grail_llvm_toolchain//tests/openssl:" + f) for f in CONF_FILES],
     outs = CONF_FILES,
     cmd = "\n".join([
-        "cp $(location @com_grail_bazel_toolchain//tests/openssl:{0}) $(location {0})".format(f)
+        "cp $(location @grail_llvm_toolchain//tests/openssl:{0}) $(location {0})".format(f)
```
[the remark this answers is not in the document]
gferon: I just did a mass rename without thinking too much. Will fix it.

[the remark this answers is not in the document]
gferon: Actually, how about renaming the original name of the repository to also match the new one with `bzlmod`, it's not really a breaking change IMO.

gferon:
> Something isn't quite right with the external_test as it seems to be running the non-bzlmod version for the bzlmod testcase.

I switched from `--experimental_enable_bzlmod` to `--enable_bzlmod` in the external tests shell script and it looks like it works now (tested locally).

jsharpe:
> > Something isn't quite right with the external_test as it seems to be running the non-bzlmod version for the bzlmod testcase.
> 
> I switched from `--experimental_enable_bzlmod` to `--enable_bzlmod` in the external tests shell script and it looks like it works now (tested locally).

That won’t work for the 5.1.0 tests..

siddharthab:
A recent change has upped the minimum supported version to 6.0.0 from 5.1.0, if it helps.

MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "grail_llvm_toolchain",
```
siddharthab, line 2: I think it will be OK to call it simply "llvm_toolchain". I am trying to get this project moved to the bazel-contrib repo, where the simpler name would make more sense.

gferon: Done! I would love to see this merged to `bazel-contrib`.

tests/scripts/run_external_tests.sh
```diff
@@ -37,14 +39,25 @@ test_args=(
...
+#   overriding rules_rust locally as its not yet released in the BCR
 "${bazel}" --bazelrc=/dev/null test "${test_args[@]}" -- \
   //foreign:pcre \
   @openssl//:libssl \
-  @rules_rust//test/unit/{native_deps,linkstamps,interleaved_cc_info}:all \
+  @rules_rust//test/unit/interleaved_cc_info:all \
   @io_bazel_rules_go//tests/core/cgo:all \
   -@io_bazel_rules_go//tests/core/cgo:cc_libs_test \
   -@io_bazel_rules_go//tests/core/cgo:external_includes_test \
+  -@io_bazel_rules_go//tests/core/cgo:oracle_convention_darwin_dylib_test \
```
siddharthab, line 59: You will also need to exclude the build target for libversioned.dylib.

siddharthab (approved):
Looks very good. Just some final nits.

toolchain/internal/configure.bzl
```diff
@@ -52,14 +62,16 @@ def llvm_register_toolchains():
         return
     arch = _arch(rctx)
 
-    (key, toolchain_root) = _host_os_arch_dict_value(rctx, "toolchain_roots")
+    if not rctx.attr.toolchain_roots:
+        toolchain_root = "@@%s_llvm//" % rctx.attr.name if BZLMOD_ENABLED else "@%s_llvm//" % rctx.attr.name
```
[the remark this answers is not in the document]
gferon: I'm not exactly sure I understand what you mean, wouldn't `bazel query` only return targets?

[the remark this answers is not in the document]
siddharthab: `bazel query` (at least before bzlmod) could also return repository configurations with the `//external` package. I can look into this later. You can go ahead and merge now.

siddharthab:
Let me merge this for you now, and I can iterate on the failing tests on my side.

siddharthab:
Thank you so much for all your hard work on this.

toolchain/extensions.bzl
jsharpe: @fmeum there is actually a real issue with this; if a transitive module requires access to the llvm_toolchain_llvm repo it can't `use_repo` on the extension as it hits this fail and so can't make the toolchain that was created by the root module visible into the transitive module where it needs to be used.

fmeum: @jsharpe Since this `fail` only applies if there is a `toolchain` tag, transitive modules should still be able to `use_extension` and `use_repo` this if they don't specify any tags. Are you seeing different behavior?

However, transitive modules of course can't know which name the root module chose for the toolchain. I would thus recommend to make `name` an optional attribute and have it default to something that transitive modules can hardcode (e.g. `llvm`). Ideally, the module would then *always* register that toolchain with some reasonable default even if the user doesn't, so that transitive modules can rely on it being present.

An alternative could be to expose a repo under a fixed name that contains a Starlark variable pointing to the first toolchain repo defined by the user. See https://github.com/bazelbuild/rules_go/blob/f03a723c49d5ec8afa0f9c74c4c3a705b37e8f40/go/private/extensions.bzl#L4-L21 for an example.

All of this depends on how you envision transitive modules using the toolchain: Should a transitive module be able to use the toolchain without the root module's cooperation? 

CC @siddharthab

PR #200, opened 2023-05-24
fix: remove use of label ctor in repo rule impl to avoid restarts

mattem:
Removes the use of the `Label` constructor used for templating with a predeclared dependency on the files as private rule attrs. This removes the need for Skyframe to restart the rule each time a new dependency is encountered during the running of the repository rule.

From running `bazel test //...` in the test directory before this change:
 
Screen Shot 2023-05-24 at 3 46 43 PM

Where as the profile after shows just the single invocation.

This is a similar issue we caught in rules_nodejs, https://github.com/bazelbuild/rules_nodejs/pull/2621 and there's further context in https://github.com/bazelbuild/bazel/issues/16162

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -189,7 +189,7 @@ def llvm_register_toolchains():
     # Convenience macro to register all generated toolchains.
     rctx.template(
         "toolchains.bzl",
-        Label("//toolchain:toolchains.bzl.tpl"),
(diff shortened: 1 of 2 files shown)


jsharpe:
@mattem this generally looks good to me, just needs fixing up so it passes CI

mattem:
🤦 missed the commit of the typo fix.

jsharpe (approved):

PR #189, opened 2023-02-14
Remove use of -undefined dynamic_lookup on darwin

keith:
This flag ignores undefined symbols at link time and forces them to be
looked up dynamically. This is incompatible with the newer fixup chains
macho format and theoretically shouldn't be necessary assuming your
dependencies are well defined.

Done in bazel's toolchain here https://github.com/bazelbuild/bazel/pull/16414

PR #202, opened 2023-06-22
Attempt to validate #189

sluongng:


sluongng:
The CI log in other PR is no longer available and I want to see what's wrong with the external test script.

Have a hard time reproducing it locally as LLVM and different repos (like absl) support for Darwin ARM64 is sketch.

sluongng:
I won't be able to iterate like this.

Ima close this for now and find an easier way to iterate

PR #211, opened 2023-09-01
Fix OSX tests

dzbarsky:
@siddharthab Sorry about missing this in the previous PR, I hadn't bothered checking if we had OSX tests

### DIFF
--- a/tests/.bazelrc
+++ b/tests/.bazelrc
@@ -1,1 +1,2 @@
 build --incompatible_enable_cc_toolchain_resolution
+build --features=-libtool
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,10 +14,10 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest]
         script: [run_tests.sh, run_external_tests.sh]
-        bazel_version: [latest, 5.1.0]
+        bazel_version: [latest, 6.0.0]
         exclude:
           - script: run_external_tests.sh
-            bazel_version: 5.1.0
+            bazel_version: 6.0.0
     steps:
     - uses: actions/checkout@v3
     - name: test
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ implementation, please let me know and I can redirect people there.
 
 ## Quickstart
 
-Minimum bazel version: **4.2.1**
+Minimum bazel version: **6.0.0**
 
 To use this toolchain, include this section in your WORKSPACE:
 ```starlark


dzbarsky:
Ah, so it looks like this works on Bazel 6.x but not 5.1.0. I believe https://github.com/bazelbuild/bazel/commit/33e7261234173fe5f4b6ce12035beeba74d7ae77 added support for the libtool feature in a 6.0pre (May 2022). How do you feel about bumping min bazel to 6.0, or would you rather hold of until 7.0 ships?

siddharthab:
6.0 for minimum supported version is fine. People can always hold off on updating to a new release of this repo, there have not been that many changes.

Will you be able to update the tests and the documentation?

dzbarsky:
Yep just kicked off a run, let's see if it works

siddharthab:
@dzbarsky Are you able to merge these changes yourself? I forgot what the settings on this repo are, and I no longer have access to them.

tests/.bazelrc
```diff
@@ -1 +1,2 @@
 build --incompatible_enable_cc_toolchain_resolution
+build --features=-libtool
```
keith, line 2: FYI there is `llvm-libtool-darwin` as well that we could use instead

dzbarsky:
Don't think I have permission to merge, I'd need write access.

tests/.bazelrc
dzbarsky: Oh, interesting. Any pointers to a comparison of the tradeoffs?

keith: if this is required for downstream users not requiring it would be the biggest benefit IMO

siddharthab: I will merge these changes now to fix the tests. @dzbarsky please consider using llvm-libtool-darwin in a separate change.

siddharthab:
> Don't think I have permission to merge, I'd need write access.

I will see if I can talk to someone with my previous employer to give me admin access again.

jsharpe:
> > Don't think I have permission to merge, I'd need write access.
> 
> I will see if I can talk to someone with my previous employer to give me admin access again.

alternatively we could consider migrating this repo to bazel-contrib? See https://bazel-contrib.github.io/SIG-rules-authors/hosting-policy.html for requirements for doing that.

siddharthab:
> > > Don't think I have permission to merge, I'd need write access.
> > 
> > 
> > I will see if I can talk to someone with my previous employer to give me admin access again.
> 
> alternatively we could consider migrating this repo to bazel-contrib? See https://bazel-contrib.github.io/SIG-rules-authors/hosting-policy.html for requirements for doing that.

Sounds good. There are 3 bazel related repos in the grailbio org, that I don't think anyone actively maintains anymore. I will see if we can get them transferred over to bazel-contrib.

tests/.bazelrc
siddharthab: I made an attempt in #229 to use llvm-libtool-darwin. It needs more work.

PR #217, opened 2023-09-14
15.0.2 and fixes /etc/os-release parsing bug

scasagrande:
- Adds clang 15.0.2
- Fixes a bug where if `/etc/os-release` contains an empty line, then we raise an index error

### DIFF
--- a/toolchain/internal/common.bzl
+++ b/toolchain/internal/common.bzl
@@ -43,6 +43,8 @@ def _linux_dist(rctx):
     info = {}
     for line in res.stdout.splitlines():
         parts = line.split("=", 1)
+        if len(parts) == 1:
+            continue
         info[parts[0]] = parts[1]
 
     distname = info["ID"].strip('\"')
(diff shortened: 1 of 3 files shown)


toolchain/internal/common.bzl
```diff
@@ -43,6 +43,8 @@ def _linux_dist(rctx):
     info = {}
     for line in res.stdout.splitlines():
         parts = line.split("=", 1)
+        if len(parts) == 1:
+            continue
```
scasagrande, line 47: In some cases, such as `manylinux2014`, you can have an `/etc/os-release` of the following format:

```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

When we're scanning the lines in this file, the line that only contains a newline character will raise an index error when executing `info[parts[0]] = parts[1]`. This change simply skips these lines, and anything that doesn't contain a '='.

siddharthab (approved):

PR #220, opened 2023-09-18
Add 16.0.5 and 16.0.6 distributions

adzenith:
I ran these commands:
```
utils/llvm_checksums.sh -g -v 16.0.4
utils/llvm_checksums.sh -g -v 16.0.5
utils/llvm_checksums.sh -g -v 16.0.6
```
and integrated the results into the distributions file. (I updated 16.0.4 because it looks like there are new distributions since this file was last updated).


siddharthab:
Thanks!

adzenith:
Thanks for the project, and for the fastest merge I've ever seen! :)

PR #223, opened 2023-09-20
Avoid calling xcrun on darwin if we don't need to

dzbarsky:
Calling xcrun can actually fail if you don't have xcode installed or haven't accepted the license/etc. If the user provides an explicit sysroot, we will never use the default one so this failure is unfortunate. Defer computing the default until we need it.


siddharthab (approved):
Thanks.

PR #230, opened 2023-10-06
Better document supports_dynamic_linker

siddharthab:
This was mentioned in #189 but did not quite surface properly.

Will also change the release notes for some recent releases.


siddharthab:
Release notes have been edited.

PR #231, opened 2023-10-06
Use rctx.read instead of cat to parse /etc/os-release

dzbarsky:
Shorter code and avoids a dependency on system `cat`


jsharpe (approved):
Thanks!

Issue #192, opened 2023-03-10
`--features=asan` doesn't work on macOS


garymm:
Building my code with `bazel build --features=asan` fails with:
`ld: file not found: external/llvm_toolchain_llvm/lib/clang/15.0.6/lib/darwin/libclang_rt.asan_osx_dynamic.dylib`

Using LLVM 15.0.6.
The file does exist. I haven't really investigated much, just filing the issue so it's common knowledge.

siddharthab:
We stopped making dylib files available in the toolchain during the build phase because the files will not be available when running or deploying your application. But perhaps we overdid it, and we can make the .dylib files in the `lib/clang/{llvm_version}/lib/darwin/` directory available.

To test, you can add a glob expression [here](https://github.com/grailbio/bazel-toolchain/blob/master/toolchain/BUILD.llvm_repo#L60) and see if it works for you. If it does, please send a PR.

keith:
For iOS builds with the standard bazel toolchain we copy the dylibs into the shipping bundle in that case, but that's easier in that setup because there's a clear place to do that piece

siddharthab:
Thanks. My original imagination here was that the osx sysroot could be a good proxy of what's available at runtime. And so if a dylib is not there, it's not available. But this looks like a special case.

timbess:
FWIW, I tested the suggested change and it works, but I ended up having to add `@llvm_toolchain//:linker-files-aarch64-darwin` as a `data` dependency of my binary so that it could find other dylibs at runtime. 
<img width="1444" alt="image" src="https://github.com/grailbio/bazel-toolchain/assets/3655689/e413debb-7c5c-40ab-a18e-94b78573d1a9">

It seems like the binary is executing in the `bazel-bin` and the dylibs aren't created in any of the places the RPATH looks for it in unless I add the linker files as a data dep.

timbess:
<img width="687" alt="image" src="https://github.com/grailbio/bazel-toolchain/assets/3655689/613e581c-81c8-46c4-bebe-07b1cc097804">
Ok so using the `_llvm` repo to grab the `:lib` target at least doesn't directly reference my toolchain triple by name, so I'll just gate this behind a select I guess for now.

PR #250, opened 2024-02-03
Fix santizers on MacOS that require dylibs in the LLVM toolchain.

timbess:
Add dylibs to the `lib` target so that santizers work properly on MacOS.

Should address https://github.com/grailbio/bazel-toolchain/issues/192.


siddharthab:
Thanks for the contribution. I am not knowledgeable enough to be confident if this is the right approach, but let's land this and see what happens.

toolchain/cc_toolchain_config.bzl
```diff
@@ -206,32 +206,22 @@ def cc_toolchain_config(
...
-            # explicitly as part of the search path to make it easy for a user
-            # to pick up something. This also makes the behavior consistent with
-            # targets when a user explicitly depends on something like
-            # libomp.dylib, which adds this directory to the search path, and would
-            # (unintentionally) lead to static linking of libraries from the
-            # toolchain.
-            link_flags.extend([
+                "-Bstatic",
+                "-lunwind",
+                "-Bdynamic",
```
fmeum, line 221: This seems to contradict the comment, wouldn't we want to link libraries from the toolchain (not the sysroot) statically instead?

keith: note that in general this is wrong for macOS since `-Bstatic` and `-Bdynamic` are not supported, but since clang is the linker they are accepted and just ignored. This results in libunwind being linked dynamically but no rpaths being set on the binary and a runtime crash similar to https://github.com/bazel-contrib/toolchains_llvm/issues/224

fmeum: @keith Should we revert or do you see a simple way to fix this?

keith: maybe a better solution for libunwind would be to just exclude it from the glob above on macOS so that the system one _has_ to be used?

keith: I think this wasn't spotted because you actually have to use something from it for it to be an issue, which is the case we hit

timbess: Sorry didn't see this until now somehow... If I remember right, I was trying to get all the sanitizers working which it wouldn't find without the dylibs in the glob, but then it would end up dynamically linking libunwind with an RPATH that referenced a relative path within the Bazel sandbox that was now gone? There was a static archive in the same directory, but ld64 would prefer the dylib unless I set `-Bstatic` after which didn't show as being dynamically linked when I ran `otool -L`, but maybe @Keith is right and I just effectively hid the error until any libunwind symbols are actually used?

Ah ok so I just replicated it. If I comment out the `-Bstatic` I get this:
image

But with the static linking I get:
image

siddharthab: I had some comments about this behaviour in the code before this change. Reverting to some of that may solve the issues here.

timbess: Seems like I'm getting `libunwind` from the actual LLVM toolchain, not the sysroot.
image

I think the issue was that it was dynamically linking to a library that was symlinked into the sandbox, but the Bazel output dir isn't in the RPATH of the final binary, so it'd fail to find `libunwind` at boot.

PR #277, opened 2024-02-26
Add missing LLVM binaries

Bo98:
And fix powerpc64le and macOS logic on these versions.


Bo98:
Sorry, took me a bit to figure out how the lint check worked. Should be good now!

siddharthab:
Thank you for the change, and also for reminding me that I need to add instructions to the README on how to set up the linter.

PR #281, opened 2024-03-05
Add support for AlmaLinux

scasagrande:
Adds support for `AlmaLinux`, which is the basis of `manylinux_2_28`

It should be treated the same as `rhel`

```
[root@beb818f8869f /]# cat /etc/os-release 
NAME="AlmaLinux"
VERSION="8.9 (Midnight Oncilla)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.9"
```

### DIFF
--- a/toolchain/internal/common.bzl
+++ b/toolchain/internal/common.bzl
@@ -34,7 +34,22 @@ def host_os_key(rctx):
     else:
         return "%s-%s-%s" % (os, version, arch)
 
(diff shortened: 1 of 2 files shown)


scasagrande:
Tested inside of a `manylinux_2_28` container, targetting 15.0.2

Example log output:

```
Fetching https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.2/clang%2Bllvm-15.0.2-x86_64-unknown-linux-gnu-rhel86.tar.xz; 242.7 MiB (35.1%) 25s
```

PR #284, opened 2024-03-12
Update remaining deps

siddharthab:


MODULE.bazel
```diff
@@ -19,6 +19,6 @@ module(
     bazel_compatibility = [">=7.0.0"],
 )
 
-bazel_dep(name = "bazel_skylib", version = "1.4.2")
-bazel_dep(name = "rules_cc", version = "0.0.8")
-bazel_dep(name = "platforms", version = "0.0.7")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
```
fmeum, line 22: Since Bzlmod uses MVS, it's generally better to *not* update dependencies, unless of course we require a bug fix or new version. Users can do so painlessly if needed and otherwise enjoy greater freedom in choosing the versions that work for them.

PR #289, opened 2024-03-13
Warn when not generating toolchain

siddharthab:


siddharthab:
Sorry, realized this is going to be too spammy. Will revert.

PR #295, opened 2024-03-14
Update compability_level in MODULE.bazel

siddharthab:


MODULE.bazel
```diff
@@ -15,7 +15,7 @@
 module(
     name = "toolchains_llvm",
     version = "0.0.0",
-    compatibility_level = 0,
+    compatibility_level = 1,
```
fmeum, line 18: Did we actually break any core API usage with recent changes? If not, I would recommend not bumping this. It certainly doesn't have to be in sync with the major version number unless we expect a large fraction of our users to be broken by an update.

siddharthab: Thanks. Will revert.

PR #302, opened 2024-03-28
Add LLVM 18 links

adzenith:
I ran `utils/llvm_checksums.sh -g -v` for 18.1.0, 18.1.1, and 18.1.2 and pasted the results into the distributions file.


rrbutani (approved):
We'll want to re-run this if/when 18.1.x binaries for `arm64-apple-darwin`/`x86_64-linux`/etc are added but I see no harm in merging this now.

Thanks!

Issue #243, opened 2024-01-03
Use LLVM toolchain in a module which can be used as both a dependency and a root module


mering:
I have a Bzlmod module "A" which I want to be able to build standalone but also be depended on by module "B".

If I define
```
bazel_dep(name = "toolchains_llvm", version = "0.10.3")
llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
llvm.toolchain(
    llvm_version = "14.0.0",
    sysroot = {
        "linux-x86_64": "chromium_x64_sysroot",
    },
    sysroot_label_map = {
        "@com_googleapis_storage_chrome_linux_amd64_sysroot//:all_files": "chromium_x64_sysroot",
    },
)
use_repo(llvm, "llvm_toolchain")
register_toolchains("@llvm_toolchain//:all")
```

in `MODULE.bazel` of both module "A" and "B", then I get the following error when compiling module "B" which depends on module "A":
```
ERROR: Traceback (most recent call last):
        File "/home/USER/.cache/bazel/_bazel_builder/395eea75460def8a11dd35795d615210/external/toolchains_llvm~override/toolchain/extensions/llvm.bzl", line 13, column 17, in _llvm_impl_
                fail("Only the root module can use the 'llvm' extension")
Error in fail: Only the root module can use the 'llvm' extension
```

But if I remove it from module "A", I can no longer build module "A" correctly standalone.

How can I define the llvm_toolchain such that I can use it in both modules "A" and "B"?

---

Based on #235.

mering:
Would it be possible to just ignore non-root tool chain calls and print a warning? This is what the Python tool chain is doing as well.

fmeum:
You can add `dev_dependency = True` to your `use_extension` and `register_toolchain` calls. Then the toolchain definition in A will be ignored when building from B. If needed, B can define its own toolchain in this way.

PR #305, opened 2024-04-02
Return, don't fail, when extension used by non-root module

scasagrande:
This fixes the following scenario:

- Two bazel modules, both of which depend on toolchains_llvm
- Both modules are bzlmod enabled with `MODULE.bzlmod` files
- Both `MODULE.bzlmod` files specify a llvm toolchain
- The second modules depends on the first via `bazel_dep`

**Expected results**

- the second module can depend on the first, and operations use the llvm toolchain that it defines

**Actual results**

- build fails with `Error in fail: Only the root module can use the 'llvm' extension`

**Resolution**

Instead of failing, we just `return`. This keeps the behaviour of only being able to actually use the `llvm` extension in the root module, but we don't error out.

A more complex but similar example is that from `rules_python` where they will still register the toolchain, but it will not be marked as the default toolchain if it is not called from the root module: https://github.com/bazelbuild/rules_python/blob/main/python/private/bzlmod/python.bzl#L102

### DIFF
--- a/toolchain/extensions/llvm.bzl
+++ b/toolchain/extensions/llvm.bzl
@@ -37,7 +37,8 @@ def _root_dict(roots, cls, name, strip_target):
 def _llvm_impl_(module_ctx):
     for mod in module_ctx.modules:
         if not mod.is_root:
-            fail("Only the root module can use the 'llvm' extension")
+            # Only the root module can use the 'llvm' extension
+            return
         toolchain_names = []
         for toolchain_attr in mod.tags.toolchain:
             name = toolchain_attr.name


rrbutani:
As per @fmeum's comment in #243, `dev_dependency` is one "solution" for this. I'm definitely open to the tack this PR takes though; the `dev_dependency` route doesn't seem ideal.

But, from a composability standpoint, doing nothing for non-root modules doesn't seem ideal either though; I see a couple of issues:
  1. rdeps of the module that uses `toolchains_llvm` will have to pull in and configure `toolchains_llvm` themselves — even if these rdeps don't themselves make direct use of a C/C++/etc toolchain or anything else from `toolchains_llvm`
  2. If I understand correctly how bzlmod extensions (with `isolate = False`) work, rdeps configuring `toolchains_llvm` will need to be careful to use the same _repo names_ as their dependencies so that `use_repo`/`register_toolchains` calls in their deps' `MODULE.bazel` files don't error

For the common use case (using `toolchains_llvm` solely via the registered C/C++ toolchain and not using other outputs like `@llvm_toolchain//:omp`, using `llvm_toolchain` as the repo name, rdeps that themselves would need/want `toolchains_llvm` anways) the above is perhaps acceptable: configuring a C/C++ toolchain is arguably something that's best decided by the top-level module anyways. (Though `dev_dependency` is maybe a better solution here; it sidesteps the repo name issue...)

However for modules that use other outputs from `llvm_toolchain` than just the C/C++ toolchain (i.e. binaries like `clang-format`, libraries like OpenMP) the first issue seems unacceptable — I should be able to properly encapsulate my module's OpenMP dep, users of my module shouldn't need to be aware of it (afaik this is not possible under a "skip non-root modules" scheme, even with `isolate = True`).

---

The least-worst solution I can come up with that addresses the above is to have `toolchains_llvm` create repositories for all modules and to handle conflicts (in repo names) by choosing the tag that's closest to the root module (as [`rules_python` does](https://github.com/bazelbuild/rules_python/blob/b9f39bf09760ff75aedd64ce99c9573c1a97068f/python/private/bzlmod/python.bzl#L136-L151)) and by warning users/nudging them towards `isolate = True`.

re: gating toolchain resolution for non-root modules: I'm not super convinced that we *need* to do this (my understanding is that toolchains registered by the root module/by modules closer to the root module have precedence in toolchain resolution; `rules_python`'s reasoning for this restriction is [here](https://github.com/bazelbuild/rules_python/blob/b9f39bf09760ff75aedd64ce99c9573c1a97068f/python/private/bzlmod/python.bzl#L103-L109)) but if we want to, it should be easy to emulate with `:all` patterns that potentially expand to nothing or by using the hub pattern that `rules_python` uses (unlike `rules_python` we don't provide a "default" toolchain in `toolchains_llvm` though).

@jsharpe @fmeum: WDYT?

scasagrande:
Yeah, using `dev_dependency=True` is totally fair for a lot of use cases. If we want to keep the `fail` call then perhaps we should update the message to suggest this solution.

Based on your analysis, I believe it means that my change shouldn't be needed for simple cases, and doesn't solve the problem for more complex ones. Is that accurate?

fmeum:
I see two separate threads here:
1. Having one "official" way of using `toolchains_llvm` in a module that is also meant to be used by other modules.
2. Supporting transitive module's needs to access LLVM toolchains without the root module's support.

For 1., I think that we should stick to and recommend `dev_dependency`: It's built right into Bazel, which means that it works uniformly for all extensions without special support, and is also self-descriptive in that you can immediately tell a certain extension tag won't do anything if the module isn't the root module.

Keeping the `fail` but having its message suggest using `dev_dependency` is thus my favorite approach within the scope of this PR.

Dealing with 2. is much more difficult, which is why I would prefer to keep the `fail` until we have properly thought through the interactions around this. It likely entails moving away from user-supplied `register_toolchains` calls, as rules_python and rules_go have done. This does add a fair share of complexity though. `isolate = True` is also not a sufficient answer as it can result in linking the same library twice in different versions.

For tools such as `clang-format`, having a module register an LLVM C++ toolchain just to access it sounds overly complicated to me. We could offer a more lightweight way to achieve this (say by not registering the `cc_toolchain`, thus avoiding the problem of reconciling toolchain requirements) or direct such modules to accept a label pointing to `clang-format` as part of their configuration instead of bringing it in themselves.

I do lack the knowledge about OpenMP to assess what a good way would be to bring it in. Does it have be kept in sync with compiler versions?

PR #312, opened 2024-04-10
Fix double slashes for sysroot = /

keith:
Ideally we could set the paths to `%sysroot%include` when we know that
`%sysroot%` ends with `/`, but bazel validates that the prefix is
actually `%sysroot%/` so we have to keep the extra slash.

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -538,7 +538,7 @@ cc_toolchain(
             # system module map.
             dir
             for dir in cxx_builtin_include_directories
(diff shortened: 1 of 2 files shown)


toolchain/internal/configure.bzl
```diff
@@ -581,8 +581,8 @@ native_binary(
         exec_dl_ext = exec_dl_ext,
     )
 
-def _is_hermetic_or_exists(rctx, path, sysroot_prefix):
```
keith, line 584: afaict this was just wrong, and we were replacing `%sysroot%` with itself

fmeum (approved):

PR #327, opened 2024-05-10
Allow explicitly specifying Linux distribution

illicitonion:
Without this, if you want to do remote execution from a non-Linux (or different Linux) platform to a Linux platform, you need to explicitly fill in the URLs attribute.

With this, you can write something like:
```
llvm_toolchain(
    name = "llvm_toolchain_linux_x86_64",
    llvm_version = "14.0.0",
    exec_os = "linux",
    exec_arch = "x86_64",
    exec_linux_distribution = "ubuntu",
    exec_linux_distribution_version = "18.04",
    sysroot = {
        "linux-x86_64": "@org_chromium_sysroot_linux_x64//:sysroot",
    },
)
```

which will work when remote building from macOS, because it won't attempt to read the Linux distribution from `/etc/os-release` which doesn't exist on macOS.

### DIFF
--- a/toolchain/internal/common.bzl
+++ b/toolchain/internal/common.bzl
@@ -93,9 +93,14 @@ def os_version_arch(rctx):
     _os = os(rctx)
     _arch = arch(rctx)
 
-    if _os == "linux" and not rctx.attr.exec_os:
-        (distname, version) = _linux_dist(rctx)
-        return distname, version, _arch
+    if _os == "linux":
+        if (rctx.attr.exec_linux_distribution == "") != (rctx.attr.exec_linux_distribution_version == ""):
+            fail("Either both or neither of linux_distribution and linux_distribution_version must be set")
+        if rctx.attr.exec_linux_distribution and rctx.attr.exec_linux_distribution_version:
+            return rctx.attr.exec_linux_distribution, rctx.attr.exec_linux_distribution_version, _arch
+        if not rctx.attr.exec_os:
+            (distname, version) = _linux_dist(rctx)
+            return distname, version, _arch
 
     return _os, "", _arch
 
--- a/toolchain/internal/repo.bzl
+++ b/toolchain/internal/repo.bzl
@@ -23,7 +23,7 @@ load(
 
 _target_pairs = ", ".join(_supported_os_arch_keys())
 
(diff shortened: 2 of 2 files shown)


toolchain/internal/common.bzl
```diff
@@ -93,9 +93,14 @@ def os_version_arch(rctx):
...
-        (distname, version) = _linux_dist(rctx)
-        return distname, version, _arch
+    if _os == "linux":
+        if (rctx.attr.exec_linux_distribution == "") != (rctx.attr.exec_linux_distribution_version == ""):
+            fail("Either both or neither of linux_distribution and linux_distribution_version must be set")
+        if rctx.attr.exec_linux_distribution and rctx.attr.exec_linux_distribution_version:
+            return rctx.attr.exec_linux_distribution, rctx.attr.exec_linux_distribution_version, _arch
+        if not rctx.attr.exec_os:
+            (distname, version) = _linux_dist(rctx)
+            return distname, version, _arch
```
rrbutani, line 103: To read `/etc/os-release` when `exec_os = "linux"` + the host platform is Linux (not just when `exec_os = None`) I think we'd want:

```suggestion
        if rctx.os.name == "linux":
            (distname, version) = _linux_dist(rctx)
            return distname, version, _arch
```

---

I don't feel strongly about whether we should do the above or change the docs on `exec_linux_distribution` to describe the existing behavior but we should probably do one or the other.

rrbutani (commented):
Thanks for the PR! This looks good to me, just a couple of nits.

---

A comment on the new attrs: in retrospect I think it would have been nicer if we'd bundled the `exec_*` override attrs into a single list attr:
```starlark
llvm_toolchain(
    ...,
    exec_platform = ["linux", "x86_64", "ubuntu", "18.04"],
)
```
```starlark
llvm_toolchain(
    ...,
    exec_platform = ["darwin", "aarch64"],
)
```

^ would let us sidestep some of the awkward permutations that are allowed (i.e. setting `linux_distribution` and `linux_distribution_version` w/o setting `exec_os` to `linux` explicitly)

(as an aside, @illicitonion: thanks for covering these in the doc strings)

But at this point it's probably not worth making a breaking change over.

YuhanunCitgez:
Having the same issue with remote builds right now :( Any idea about priority on this one?

PR #339, opened 2024-06-12
Include OpenMP shared libraries

jsharpe:
Include the OpenMP shared libraries that are needed if you compile with `-fopenmp` or OpenMP target offload enabled.

### DIFF
--- a/toolchain/BUILD.llvm_repo
+++ b/toolchain/BUILD.llvm_repo
@@ -62,6 +62,10 @@ filegroup(
             "lib/**/lib*.a",
             "lib/clang/*/lib/**/*.a",
             "lib/clang/*/lib/**/*.dylib",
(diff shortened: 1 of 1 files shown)


rrbutani (approved):
Looks good!

If it isn't too much trouble would you mind adding a test for this? (`cc_binary` with `-fopenmp` that fails to build if these `.so`s aren't made available)

jsharpe:
Sure, I'll look to putting a test together for this. IIRC where we use this may also need the addition of a `-B` flag to properly find the bundler executables from the toolchain (we currently set this via .bazelrc) but this test will highlight whether this is required for a general case.

thomasbaekkegaard (commented):
FYI: `lib/*omp*` -files are not included in LLVM-19.* distributions.

PR #154, opened 2022-08-19
WIP: VERY EARLY DRAFT! WIP WIP WIP!!!! - Add clang-cl windows cross compilation

GregBowyer:
…

Hi, this will be a little insane and a little strange, but I would
like to get a read on the following as I refine it towards a potential
addition to these rules.

The following code allows for cross compiling windows (with MSVC ABI
rather than GNU, and I hope to adapt it to cope with clang on windows
too.

If this is something wanted I can polish these rules further rather than
fork and adapt.

This is very much an initial read, so there are some hacks, a hard-coded
path I could use help with and probably constructions that I am sure we
want to change.

As an example of this running (the binary is bytewise equivalent to the windows produced version) is as follows.

https://youtu.be/v54e1pSMUbw

PR #342, opened 2024-06-17
feat: windows toolchain for running clang tools

peakschris:
This is a start on windows toolchain support (https://github.com/bazel-contrib/toolchains_llvm/issues/4). Enough is in place for the distribution to be downloaded, extracted, and tools made available through symlinks and a build file.

working examples:
bazel run @llvm_toolchain_llvm//:bin/clang-tidy.exe
bazel run @llvm_toolchain_llvm//:bin/clang-format.exe

llvm 18+ is required.

Comments, suggestions, help all very welcome

### DIFF
--- a/toolchain/BUILD.llvm_repo_windows
+++ b/toolchain/BUILD.llvm_repo_windows
@@ -1,0 +1,140 @@
+# Copyright 2021 The Bazel Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+# Some targets may need to directly depend on these files.
+exports_files(glob([
+    "bin/*",
+    "lib/*",
+    "include/*",
+]))
+
+## LLVM toolchain files
+
+filegroup(
+    name = "clang",
+    srcs = [
+        "bin/clang.exe",
+        "bin/clang++.exe",
+        "bin/clang-cpp.exe",
+    ],
+)
+
+filegroup(
+    name = "ld",
+    srcs = glob(["bin/ld.lld.exe", "bin/ld64.lld.exe"]),
+)
+
+filegroup(
+    name = "include",
+    srcs = glob([
+        "include/**/c++/**",
+        "lib/clang/*/include/**",
+    ]),
+)
+
+filegroup(
+  name = "all_includes",
+  srcs = glob(["include/**"]),
+)
+
+filegroup(
+    name = "bin",
+    srcs = glob(["bin/**"]),
+)
+
+filegroup(
+    name = "lib",
+    srcs = glob(
+        [
+            "lib/**/lib*.a",
+            "lib/clang/*/lib/**/*.a",
+            "lib/clang/*/lib/**/*.dylib",
+            # clang_rt.*.o supply crtbegin and crtend sections.
+            "lib/**/clang_rt.*.o",
+        ],
+        exclude = [
+            "lib/libLLVM*.a",
+            "lib/libclang*.a",
+            "lib/liblld*.a",
+        ],
+    ),
+    # Include the .dylib files in the linker sandbox even though they will
+    # not be available at runtime to allow sanitizers to work locally.
+    # Any library linked from the toolchain to be released should be linked statically.
+)
+
+filegroup(
+    name = "ar",
+    srcs = ["bin/llvm-ar.exe"],
+)
+
+filegroup(
+    name = "as",
+    srcs = [
+        "bin/clang.exe",
+        "bin/llvm-as.exe",
+    ],
+)
+
+filegroup(
+    name = "nm",
+    srcs = ["bin/llvm-nm.exe"],
+)
+
+filegroup(
+    name = "objcopy",
+    srcs = ["bin/llvm-objcopy.exe"],
+)
+
+filegroup(
+    name = "objdump",
+    srcs = ["bin/llvm-objdump.exe"],
+)
+
+filegroup(
+    name = "profdata",
+    srcs = ["bin/llvm-profdata.exe"],
+)
+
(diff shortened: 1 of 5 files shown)


toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+]))
+
+## LLVM toolchain files
+
+filegroup(
+    name = "clang",
+    srcs = [
+        "bin/clang.exe",
+        "bin/clang++.exe",
+        "bin/clang-cpp.exe",
```
rrbutani, line 31: I think we may want (need?) to include `clang-cl.exe` for Windows

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+    ],
+)
+
+filegroup(
+    name = "ld",
+    srcs = glob(["bin/ld.lld.exe", "bin/ld64.lld.exe"]),
```
rrbutani, line 37: I think we want `lld-link.exe` for Windows

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+
+filegroup(
+    name = "lib",
+    srcs = glob(
+        [
+            "lib/**/lib*.a",
+            "lib/clang/*/lib/**/*.a",
+            "lib/clang/*/lib/**/*.dylib",
+            # clang_rt.*.o supply crtbegin and crtend sections.
+            "lib/**/clang_rt.*.o",
```
rrbutani, line 66: I believe these are `.lib`s for Windows (i.e. `clang_rt.builtins-x86_64.lib`).

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+filegroup(
+    name = "bin",
+    srcs = glob(["bin/**"]),
+)
+
+filegroup(
+    name = "lib",
+    srcs = glob(
+        [
+            "lib/**/lib*.a",
```
rrbutani, line 62: This will not match anything for the Windows distribution (`.lib` instead of `.a`) though that may be okay – I'm not actually sure what this is supposed to pull in that's not covered by the below globs.

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+        ],
+        exclude = [
+            "lib/libLLVM*.a",
+            "lib/libclang*.a",
+            "lib/liblld*.a",
+        ],
+    ),
+    # Include the .dylib files in the linker sandbox even though they will
+    # not be available at runtime to allow sanitizers to work locally.
+    # Any library linked from the toolchain to be released should be linked statically.
```
fmeum, line 76:
```suggestion
```

toolchain/internal/configure.bzl
```diff
@@ -545,15 +543,21 @@ cc_toolchain(
         major_llvm_version = major_llvm_version,
     )
 
+def _extension(os):
+    if (os == "windows"):
```
fmeum, line 547:
```suggestion
    if os == "windows":
```

toolchain/internal/configure.bzl
```diff
@@ -570,6 +574,7 @@ cc_import(
 
     tool_target_strs = []
     for name in _aliased_tools:
+        name = name+ext
```
fmeum, line 577:
```suggestion
        name = name + ext
```

toolchain/internal/release_name.bzl
```diff
@@ -46,6 +46,16 @@ def _darwin(llvm_version, arch):
     )
 
 def _windows(llvm_version, arch):
+    major_llvm_version = _major_llvm_version(llvm_version)
+    if (major_llvm_version >= 18 and arch.endswith("64")):
```
fmeum, line 50:
```suggestion
    if major_llvm_version >= 18 and arch.endswith("64"):
```

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
+# Copyright 2021 The Bazel Authors.
```
rrbutani, line 1: WDYT about unifying this with `BUILD.llvm_repo`? It seems like the discrepancies are mostly just in the file extensions (`.exe`, `.lib` for Windows).

Maybe something like:
```starlark
is_windows = {is_windows} # provided by `llvm_repo`
executable_extension = ".exe" if is_windows else ""

filegroup(
  name = "clang",
  srcs = [
    "bin/" + e + executable_extension for e in [
      "clang", "clang++", "clang-cpp", "clang-cl",
    ]
  ],
)

# ...
```

toolchain/BUILD.llvm_repo_windows
```diff
@@ -0,0 +1,140 @@
...
+
+filegroup(
+  name = "all_includes",
+  srcs = glob(["include/**"]),
+)
+
+filegroup(
+    name = "bin",
+    srcs = glob(["bin/**"]),
+)
```
rrbutani, line 56: I'm not very familiar with Windows but IIUC the binaries require the DLLs in `bin/` in the Windows release tarball to function; I think we need to add them to `bin` and `clang`, `as`, `ar`, etc?

rrbutani (commented):
@peakschris Thanks for taking a stab at this.

I'm not super familiar with Windows C/C++ development but I left some notes.

If you haven't come across it already, #154 may be of interest to you; it uses `cc_toolchain_config` from `@bazel_tools//tools/cpp:windows_cc_toolchain_config.bzl` which we'll probably also want to eventually do.

toolchain/BUILD.llvm_repo_windows
peakschris: I don't need the DLLs as I already have them installed via MSVC on our systems. But for others without MSVC this is a good point. Have a look at my fix, as I don't know how these filegroups will be consumed downstream I'm not sure if the DLLs should go in srcs or data or simply allow downstream to consume separately from the dlls filegroup.

peakschris:
> If you haven't come across it already, #154 may be of interest to you; it uses `cc_toolchain_config` from `@bazel_tools//tools/cpp:windows_cc_toolchain_config.bzl` which we'll probably also want to eventually do.

This is a great reference, thank you! I don't understand cross compilation, but could give a go at integrating the windows_cc_toolchain_config to a level where it works on a windows machine. I'd like to do this as a followup PR if that's OK. Maybe someone else could take on the cross compilation aspect, I can see that would be hugely valuable for CI runners.

bomkvilt:
Hi, is the PR dead in the end?

PR #363, opened 2024-08-01
Add Publish to BCR setup

fmeum:


.bcr/metadata.template.json
```diff
@@ -0,0 +1,23 @@
...
+      "github": "fmeum",
+      "name": "Fabian Meumertzheim"
+    },
+    {
+      "email": "james.sharpe@zenotech.com",
+      "github": "jsharpe",
+      "name": "James Sharpe"
+    },
+    {
+      "email": "rrbutani@users.noreply.github.com",
```
fmeum, line 15: @rrbutani I can add your real email here, I just didn't know whether you would be okay with it being so public.

jsharpe (approved):

PR #371, opened 2024-08-09
Allow user provided platform constraints

jkurland-roku:
An implementation of the fix suggested in https://github.com/bazel-contrib/toolchains_llvm/issues/361 to allow users to specify additional platform constraints for each toolchain.

My personal use case was building some targets with musl and the toolchains here were interferring.


toolchain/internal/repo.bzl
```diff
@@ -271,6 +271,14 @@ llvm_config_attrs.update({
         default = False,
         doc = "Use absolute paths in the toolchain. Avoids sandbox overhead.",
     ),
+    "extra_exec_compatible_with": attr.string_dict(
+        mandatory = False,
+        doc = "Extra constraints to be added to exec_compatible_with for each target, with json encoded values to get a Dict[string, List[string]]",
```
fmeum, line 276:
```suggestion
        doc = "Extra constraints to be added to exec_compatible_with for each target, with json encoded values to get a dict[str, list[str]]",
```
following https://docs.google.com/document/d/1Sid7EAbBd_w_T7D94Li_f_bK3zMTztFbzIMvcpzo1wY/edit

toolchain/internal/repo.bzl
```diff
@@ -271,6 +271,14 @@ llvm_config_attrs.update({
...
+    ),
+    "extra_target_compatible_with": attr.string_dict(
+        mandatory = False,
+        doc = "Extra constraints to be added to target_compatible_with for each target, with json encoded values to get a Dict[string, List[string]]",
```
fmeum, line 280:
```suggestion
        doc = "Extra constraints to be added to target_compatible_with for each target, with json encoded values to get a dict[str, list[str]]",
```

toolchain/extensions/llvm.bzl
```diff
@@ -34,6 +34,23 @@ def _root_dict(roots, cls, name, strip_target):
 
     return res
 
+def _constraint_dict(tags, name):
+    cons = {}
```
fmeum, line 38: `cons` has so many meanings, lets call this `constraints`.

toolchain/internal/repo.bzl
```diff
@@ -271,6 +271,14 @@ llvm_config_attrs.update({
         default = False,
         doc = "Use absolute paths in the toolchain. Avoids sandbox overhead.",
     ),
+    "extra_exec_compatible_with": attr.string_dict(
```
fmeum, line 274: Could this be an `attr.string_list_dict` instead? Then we wouldn't have to use JSON encoding at all.

fmeum (commented):
Thanks for picking this up, looks pretty good overall already.

toolchain/internal/repo.bzl
jkurland-roku: absolutely, can't believe I've never noticed that one before, thanks

fmeum:
Could you add a test usage of these new attributes to https://github.com/bazel-contrib/toolchains_llvm/blob/master/tests/MODULE.bazel? You could add synthetic `constraint_setting`s with default values so that the default platform matches them. I would just like to get more Starlark coverage.

jkurland-roku:
Yeah, good idea. I want to do a platform transition to actually test that everything works. Is it ok if I add a bazel_dep to https://github.com/aspect-build/bazel-lib?

jkurland-roku:
I've added a test, the only issue I think is that it causes another llvm toolchain to be downloaded and extracted (or just extracted if there's a repository cache).

tests/MODULE.bazel
```diff
@@ -66,10 +66,26 @@ llvm.toolchain(
...
 register_toolchains("@llvm_toolchain//:all")
 
+llvm.toolchain(
+    name = "llvm_toolchain_with_define",
+    llvm_versions = LLVM_VERSIONS,
+    cxx_flags = {"": ["-DTEST_DEFINE", "-stdlib=libc++"]},
+)
+llvm.extra_target_compatible_with(
+    name = "llvm_toolchain_with_define",
+    constraints = ["@//:yes_test_define"],
```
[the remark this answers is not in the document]
jkurland-roku: leave one relative to test both?

jkurland-roku:
Looks like the ubuntu failures are when bzlmod is disabled which makes sense and I'll look to fix that. The macos failures might be to do with the cxxflags, maybe there's another mechanism to detect which toolchain was selected

jkurland-roku:
I think those changes should fix both issues (although not sure about macos, I don't really know anything about macos). I've tested with bzlmod disabled so at least that should work

jkurland-roku:
So I think the issue is that on the platforms which only support up to llvm 13.0.0 the c++20 toolchain fails to build. So I'm thinking that maybe I get rid of the current c++20 check and instead make a new platform for zlib supporting linker (a better name would be helpful) and then switch to llvm 13 automatically based on that. So replacing the "extra_toolchain" arg in the CI with "--host_platform=@//:archlinux_docker" for example.

What do you think?

fmeum:
That's one approach. You could also try to disable the test for these platforms. Generic starlark changes that work on two common platforms are very unlikely to break on untested platforms.

jkurland-roku:
I've made this test manual and now only run it when the toolchain isn't overriden

fmeum (approved):

fmeum:
Thanks for carrying this through!

JKurland:
No worries, thanks for the help

PR #380, opened 2024-08-27
Always use LLD on darwin

dzbarsky:
It sounds like a few folks have tried the `-fuse-ld=lld` option and it works for them, so let's try to roll it out more broadly.


dzbarsky:
@rrbutani are you the right person to take a look at this? I tried to keep the cmd lines generated the same as we did before this change (when running with `--linkopt=-fuseld=ld64.lld` with the exception of the link sections, but perhaps we can change some of the flags for non-xcompile to the lld variants and simplify things further?

toolchain/cc_toolchain_config.bzl
```diff
@@ -228,7 +210,19 @@ def cc_toolchain_config(
                 "-Bdynamic",
                 "-L{}lib".format(toolchain_path_prefix),
             ])
-
+        else:
+            # For single-platform builds, we can statically link the bundled
```
dzbarsky, line 214: this comment is preexisting but I don't think it's correct?

rrbutani: I think it actually might be right because of this bit here:

https://github.com/bazel-contrib/toolchains_llvm/blob/6edc43be7e52d9f44bc842a9956bd40213b972b1/toolchain/cc_toolchain_config.bzl#L188-L189

i.e. `builtin-libc++` implies `not is_xcompile`, and this arm has `exec_os != "darwin"` implying `linux (exec) -> linux (target)`

rrbutani: definitely no objection to rewording that comment though

rrbutani (approved):
Apologies for the delay — unfortunately I haven't really kept up with the state of `lld` on darwin and no longer have a machine to test this on.

However, everything in this PR looks good to me. Appreciate the cleanup as well.

garymm:
Can add this to the PR description:
`Fixes: https://github.com/bazel-contrib/toolchains_llvm/issues/212`.

would be great to see this merged!

fmeum:
@dzbarsky Happy to merge after conflicts have been resolved

PR #390, opened 2024-09-17
feat: cross-compile to windows msvc

f0rmiga:
So, I had this code on my computer for a long time. I found it and wanted to share it with the community. It's not ready for a thorough review, but I wanted to spark a discussion about whether such a thing is desirable.

My motivation is simple: I don't want to have a Windows machine, much less one for development, but I want to distribute my software to run on Windows machines.

To accomplish this, I untangled some code around cross-compilation, which should also pave the way to cross-compile targeting macOS.

If we don't make any assumptions about whether we are cross-compiling, we can have a single pipeline for the targets. E.g. exec Linux or macOS targeting Linux should have no difference in which tools are used to compile and link.
Notice that my code only uses the lld already in the downloaded toolchain.

One way to obtain the Windows sysroot is by using https://github.com/Jake-Shadle/xwin. At this point, I didn't give the search paths much thought and instead hard-coded the directories I had on my sysroot.

A very important piece of this puzzle is adding the `windows_cc_toolchain_config.bzl` file with a few fixes you can see [here](https://github.com/bazel-contrib/toolchains_llvm/commit/bb33d01e6da398ee229151cafc213971aafd1912) and [here](https://github.com/bazel-contrib/toolchains_llvm/commit/8ca849f49381de0a348be8ece542408ac329cc03). What I had to change makes me believe that nobody has ever tried to set a Windows sysroot using msvc under Bazel. There's no way the original `windows_cc_toolchain_config.bzl` could have worked with a sysroot. It's also possible that I am missing something...

I hope to get some feedback, and I'm happy to answer any questions you may have.

### DIFF
--- a/platforms/BUILD.bazel
+++ b/platforms/BUILD.bazel
@@ -45,3 +45,11 @@ platform(
         "@platforms//cpu:aarch64",
     ],
 )
+
+platform(
+    name = "windows-msvc-x86_64",
+    constraint_values = [
+        "@platforms//os:windows",
+        "@platforms//cpu:x86_64",
+    ],
+)
--- a/toolchain/BUILD.llvm_repo
+++ b/toolchain/BUILD.llvm_repo
@@ -32,6 +32,7 @@ filegroup(
     srcs = [
         "bin/clang",
         "bin/clang++",
+        "bin/clang-cl",
         "bin/clang-cpp",
     ],
 )
@@ -41,6 +42,7 @@ filegroup(
     srcs = [
         "bin/ld.lld",
         "bin/ld64.lld",
+        "bin/lld-link",
     ],
 )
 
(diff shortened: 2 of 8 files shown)


f0rmiga:
@fmeum, any thoughts?

fmeum:
I generally like this a lot and can review when you think it's ready. We should strive to make the change in a way that somewhat obviously doesn't change command lines for non-Windows target platforms.

kkass:
I was about to put up a very similar implementation.  With a couple key differences.

1. I am supporting x86_64 and aarch64 builds.
2. I am not embedding windows_cc_toolchain_config, but loading it directly from rules_cc
3. I am not using the winsysroot style.  Instead I have separate crt and sdk directories (I also use xwin)
4. I have split the unix and windows cc_toolchain_config functions (it seems less confusing that way)
5. I am not creating a new cc_wrapper for clang-cl. (This is not needed since clang-cl is already a symlink.)

I would be interested in combining efforts.

Issue #345, opened 2024-06-20
Use deb packages as binary sources


layus:
I wonder if there is a way to use deb packages hosted on https://apt.llvm.org/ as binary source. While binary packages are provided by volunteers hand there is uneven support for systems, the apt repo contains releases for each llvm version and several ubuntu versions.

Has someone ever tried that ? How hard would it be to redirect toolchains_llvm to a local repository_rule containting the unpacked and merged packages ?

zaucy:
I'm also interesting in this, but I haven't tried it yet.

We could possibly use these deb packages as well http://ftp.debian.org/debian/pool/main/l/llvm-toolchain-18/

mering:
Should no longer be needed with #392

PR #392, opened 2024-09-19
Add LLVM 19

mering:
The filenames changed significantly, so I had to  add new logic. On the other hand, the file names are much easier to construct now, especially on Linux where there are no longer distro-specific variants as non-hermetic dynamic library dependencies like [`libtinfo5` have been removed](https://github.com/llvm/llvm-project/pull/93889).


zaucy:
> there are no longer distro-specific variants

Glad to hear it! I tried out your branch on my projects and it works great.

#345 should be unnecessary if this gets merged

fmeum (approved):
The new scheme is so much nicer!

mering:
@fmeum For when is the next release including this PR planned?

fmeum:
@mering I cut a new release

PR #393, opened 2024-09-21
Fix release notes template tag.

sputt:
The existing release notes template block can't be pasted into a MODULE.bazel because it generates an invalid semver.  Strip the leading "v".

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -16,8 +16,9 @@ sed -i.bak "s/0.0.0/${tag}/" MODULE.bazel && git add MODULE.bazel && git commit 
 git archive --format=tar --prefix="${prefix}/" HEAD | gzip >"${archive}"
 sha=$(shasum -a 256 "${archive}" | cut -f1 -d' ')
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release_prep.sh
```diff
@@ -16,8 +16,9 @@ sed -i.bak "s/0.0.0/${tag}/" MODULE.bazel && git add MODULE.bazel && git commit
 git archive --format=tar --prefix="${prefix}/" HEAD | gzip >"${archive}"
 sha=$(shasum -a 256 "${archive}" | cut -f1 -d' ')
 
+# Strip leading "v" from the tag if present.
 sed \
-  -e "s/{tag}/${tag}/g" \
+  -e "s/{tag}/${tag#v}/g" \
```
fmeum, line 21: Can we introduce a new `{version}` variable for this? The tag does include the `v`, so stripping it here would be confusing.

sputt: Good call, the release link for archive_override would have been wrong.  Fixed

fmeum (approved):

PR #400, opened 2024-10-11
Add dynamic-stdc++

keith:
Depending on your project statically linking libstdc++ from your sysroot
might not be viable. For example if you rely on other shared libraries
that link the shared version, linking the static version to a binary
might result in ODR violations. This new `dynamic-stdc++` sets the
compile flags correctly but doesn't force the static version, so the
sysroot's shared version will be preferred.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -273,6 +273,10 @@ def cc_toolchain_config(
             "-std=" + cxx_standard,
             "-stdlib=libstdc++",
         ]
+
+        link_flags.extend([
+            "-lstdc++",
+        ])
     elif stdlib == "stdc++":
         cxx_flags = [
             "-std=" + cxx_standard,
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -268,6 +268,11 @@ def cc_toolchain_config(
             "-l:c++.a",
             "-l:c++abi.a",
         ])
+    elif stdlib == "system-stdc++":
+        cxx_flags = [
+            "-std=" + cxx_standard,
+            "-stdlib=libstdc++",
+        ]
     elif stdlib == "stdc++":
         cxx_flags = [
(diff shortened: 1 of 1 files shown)


toolchain/cc_toolchain_config.bzl
```diff
@@ -244,6 +244,15 @@ def cc_toolchain_config(
             "-l:c++.a",
             "-l:c++abi.a",
         ])
+    elif stdlib == "system-stdc++":
```
keith, line 247: we could use `builtin-` like the other one but i read that as "builtin to this toolchain" not "builtin to the sysroot", no pref tho

keith:
i think i need diff flag here, in my testing it's still using the static version

keith:
nvm, this does work as long as you have a valid libstdc++.so in the sysroot

toolchain/cc_toolchain_config.bzl
fmeum: Maybe just call it `sysroot-stdc++`? `/` is also a sysroot, just not a hermetic one.

fmeum (approved):
LGTM modulo naming

toolchain/cc_toolchain_config.bzl
jmillikin: It seems like the difference from current behavior isn't the location of `libstdc++` (both come from the sysroot), but that dynamic linking is allowed.

I would be fairly confused if setting `libc = "stdc++"` and `libc = "sysroot-stdc++"` implied a difference in library file extension (`.a` vs `.so`) but not library path.

fmeum: That's a good point. `dynamic-stdc++` then?

keith: renamed to dynamic-stdc++

fmeum (approved):

keith:
ci fix: https://github.com/bazel-contrib/toolchains_llvm/pull/427

PR #401, opened 2024-10-11
ci: update `libtinfo5` installation for Ubuntu 24.04

rrbutani:
As per https://github.com/actions/runner-images/issues/10636 `ubuntu-latest` is now Ubuntu 24.04 instead of 22.04 which [does not] have `libtinfo5` in the default PPAs.

[does not]: https://packages.ubuntu.com/search?keywords=libtinfo

Unfortunately all of the LLVM binary releases for Ubuntu still target 18.04 ~~and contain binaries linked against `libtinfo5` — bumping the version of LLVM we use in our tests' toolchains will not solve this problem~~ (see below).

Pinning CI to use Ubuntu 22.04 is an option; however this PR installs `libtinfo5` on the Ubuntu 24.04 runners manually instead.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,10 +32,9 @@ jobs:
         bzlmod: [true, false]
     runs-on: ${{ matrix.os }}
     steps:
-      - if: startsWith(matrix.os, 'ubuntu')
-        name: Install APT packages
-        run: sudo apt-get -y install libtinfo5
       - uses: actions/checkout@v4
+      - if: startsWith(matrix.os, 'ubuntu')
+        run: tests/scripts/ubuntu_install_libtinfo.sh
       - name: Test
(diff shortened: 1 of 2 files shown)


keith:
if you want to not install it we just need to use the 19.x train of llvm, which no longer has a dep on this lib

rrbutani:
> if you want to not install it we just need to use the 19.x train of llvm, which no longer has a dep on this lib

Good catch; forgot about that change.

I think it makes sense for us to still test some of the older LLVM toolchain versions in our test suite but I don't feel strongly about this.

I'm going to leave this change as is for now to get CI unbroken but feel free to revert + bump LLVM in the tests in a follow up.

keith:
++ on testing old stuff, just figured that might be easier, but looks like you already got something working so nvm!

PR #405, opened 2024-10-17
Support WebAssembly target platforms `wasm{32,64}-unknown-unknown`

jmillikin:


fmeum (commented):
This generally looks good to me. 

Could you add a test that builds a simple binary for wasm using a transition? Ideally that would also include adding a usage of `libclang_rt`.

jmillikin:
Tests added:
* `wasm32_strlen_libc` which builds a 32-bit `.wasm` binaries with a `libc.a` dependency, and transitively requires `wasm32-unknown-unknown/libclang_rt.builtins.a`.
* `wasm64_strlen_nolibc` which builds a 64-bit `.wasm` binary with no libc dependency (since 64-bit WebAssembly is still experimental, and therefore WASI SDK doesn't exist currently).
* `wasm32_strlen_nolibc` which is the same code as ^^ but 32-bit, so `wasm-dis` can be used to manually verify the expected difference in ABI.

fmeum:
Thanks for adding tests! CI is failing now though.

jmillikin:
CI appears to be failing for infrastructure reasons. I'll try poking at it a bit, but it might require a bit of refactoring to the test scripts.

---

container_test (debian, true):
```
/root/.cache/[...]: write (No space left on device)
```

container_test (linux_sysroot, true)
```
bin/wasm-ld: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
```

test (macos-latest, latest, false)
```
ERROR: /Users/runner/work/toolchains_llvm/toolchains_llvm/tests/BUILD.bazel:209:10: Compiling wasm_strlen.c failed: (Exit 1): DUMMY_GCC_TOOL failed: error executing CppCompile command (from target //:wasm_strlen) external/rules_rust/rust/private/dummy_cc_toolchain/DUMMY_GCC_TOOL -MD -MF bazel-out/darwin_arm64-fastbuild-ST-ee9519131230/bin/_objs/wasm_strlen/wasm_strlen.d ... (remaining 18 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
src/main/tools/process-wrapper-legacy.cc:80: "execvp(external/rules_rust/rust/private/dummy_cc_toolchain/DUMMY_GCC_TOOL, ...)": No such file or directory
```

jmillikin:
Updated tests so that WebAssembly targets get their own directory, and the test script runs `bazel clean --expunge` before trying to build them -- this is intended as a mitigation to avoid running out of disk space.

Also updated `tests/WORKSPACE` to match `tests/MODULE.bazel`, and verified non-bzlmod via the following commands on my local workstation (Linux, Ubuntu 20.04)
```
bazel-7.0.0 test --enable_bzlmod=false //wasm:all
bazel-7.0.0 test --enable_bzlmod=true //wasm:all
bazel-7.3.2 test --enable_bzlmod=false //wasm:all
bazel-7.3.2 test --enable_bzlmod=true //wasm:all
```

jmillikin:
More test updates:
* Add `libxml2` to the `apt-get` setup lines in the Debian-based containers.
* Removed `-Wl,-v` from the linker options for Wasm tests to work around a bug in `wasm-ld`, and filed https://github.com/llvm/llvm-project/issues/112836 for upstream.

fmeum (approved):
Thanks, this is a great addition!

PR #409, opened 2024-10-29
Make sure that only expands existing response files

ormandi:
Otherwise the parameter is more than likely a prefix of an `@rpath` so it is probably better off leaving as it is.

This simply ports the following fix: https://github.com/bazelbuild/bazel/pull/13148

Resolves bazel-contrib/toolchains_llvm#408

### DIFF
--- a/toolchain/osx_cc_wrapper.sh.tpl
+++ b/toolchain/osx_cc_wrapper.sh.tpl
@@ -86,7 +86,7 @@ function sanitize_option() {
 
 cmd=()
 for ((i = 0; i <= $#; i++)); do
-  if [[ ${!i} == @* && -r "${i:1}" ]]; then
+  if [[ ${!i} == @* && -r ${i:1} ]]; then
     while IFS= read -r opt; do
       if [[ ${opt} == "-fuse-ld=ld64.lld" ]]; then
         cmd+=("-fuse-ld=lld")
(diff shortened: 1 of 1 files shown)


fmeum (approved):

ormandi:
@fmeum - Thank you for the approval. Had the formatting issue auto fixed. Please take another look.

toolchain/osx_cc_wrapper.sh.tpl
```diff
@@ -86,7 +86,7 @@ function sanitize_option() {
 
 cmd=()
 for ((i = 0; i <= $#; i++)); do
-  if [[ ${!i} == @* ]]; then
+  if [[ ${!i} == @* && -r ${i:1} ]]; then
```
fmeum, line 89: This would actually be safer with double quotes as arguments can contain spaces. What kind of formatting issue did you see?

ormandi: Agree, I like the original better and in fact that is exactly what is implemented in https://github.com/bazelbuild/bazel/pull/13148.

The formatting issue was raised by `trunk fmt` in the CI pipeline of `bazel-contrib/toolchains_llvm` which was marked as auto correctable. Please see it here: https://github.com/bazel-contrib/toolchains_llvm/pull/409/checks?check_run_id=32219399947 I don't use this tool, but for this reason I had downloaded it, and had it auto fix the formatting issue which resulted in the second commit.

Do you want me to revert this? Please let me know. In my opinion it would be better off to return to the original version. How could I bypass the above issue? Or this is not something that prevents merging?

fmeum: Yes, please revert to the version with quotes. If that fails CI, I will investigate how we can fix that.

ormandi: Done. PTAL

PR #411, opened 2024-10-31
Fix shfmt config to disable Bash code simplification

thirtyseven:
I'm @ormandi's colleague and was following #409. It seems the formatter wanting to remove quotes is due to Trunk by default using the `-s` flag for `shfmt` to simplify code. Either there is a bug in `shfmt` that makes it overly aggressive in removing quotes or what it's doing is actually correct, but I'm not enough of a shell expert to tell.

This PR sidesteps the issue by customizes the Trunk config to avoid passing `-s`.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -44,6 +44,18 @@ lint:
       version_command:
         parse_regex: ${semver}
         run: buildifier --version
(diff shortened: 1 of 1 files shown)


ormandi:
Thanks @thirtyseven! This is great!

PR #415, opened 2024-11-07
fix: Skip cxx_builtin_include_directories filtering when remote

honnix:
It is not possible to determine directory existence remotely, so for remote execution the best option is not to filter. This PR addresses https://github.com/bazel-contrib/toolchains_llvm/pull/280/files/1a11e9f0f58f89789b5cd0a389efc7e321e33b3b#r1830121103

### DIFF
--- a/toolchain/internal/common.bzl
+++ b/toolchain/internal/common.bzl
@@ -115,7 +115,9 @@ def os(rctx):
             return name
         else:
             fail("Unsupported value for exec_os: %s" % name)
+    return os_from_rctx(rctx)
 
+def os_from_rctx(rctx):
     name = rctx.os.name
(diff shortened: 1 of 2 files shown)


toolchain/internal/common.bzl
```diff
@@ -115,7 +115,9 @@ def os(rctx):
             return name
         else:
             fail("Unsupported value for exec_os: %s" % name)
+    return os_from_rctx(rctx)
 
+def os_from_rctx(rctx):
```
honnix, line 120: 🙈 sorry for the bad name, I'm all ears.

fmeum (approved):

fmeum:
Thanks!

PR #427, opened 2024-12-20
[CI] Fix for bazel 8.x

keith:


keith:
I changed 7.0.0 here to be 7.x, since 7.0.0 didn't have the --enable_workspace flag, we could probably fix that but i don't really think it's worth it. That requires changing the required statuses since they have the version name in them. i don't have access to do that

fmeum (approved):

PR #430, opened 2025-01-06
cc_wrapper: Spill arguments back to a response file

ParkMyCar:
Fixes https://github.com/bazel-contrib/toolchains_llvm/issues/421

This PR tweaks the `*cc_wraper.sh` scripts to spill command line arguments from response files, back into a file after sanitizing. Previously they were expanded into the `cmd` variable which would then get specified on the command line, possibly leading to an overflow.

### DIFF
--- a/toolchain/cc_wrapper.sh.tpl
+++ b/toolchain/cc_wrapper.sh.tpl
@@ -60,14 +60,19 @@ function sanitize_option() {
 
 cmd=()
 for ((i = 0; i <= $#; i++)); do
-  if [[ ${!i} == @* ]]; then
+  if [[ ${!i} == @* && -r "${i:1}" ]]; then
+    # Create a temporary file that we'll spill sanitized options to since they
+    # were originally read from a response file.
+    temp_file=$(mktemp)
+
     while IFS= read -r opt; do
       opt="$(
         set -e
         sanitize_option "${opt}"
       )"
-      cmd+=("${opt}")
+      echo "${opt}" >> "${temp_file}"
     done <"${!i:1}"
+    mv "${temp_file}" "${!i:1}"
   else
     opt="$(
       set -e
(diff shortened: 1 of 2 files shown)


toolchain/cc_wrapper.sh.tpl
```diff
@@ -60,14 +60,19 @@ function sanitize_option() {
 
 cmd=()
 for ((i = 0; i <= $#; i++)); do
-  if [[ ${!i} == @* ]]; then
+  if [[ ${!i} == @* && -r "${i:1}" ]]; then
+    # Create a temporary file that we'll spill sanitized options to since they
+    # were originally read from a response file.
+    temp_file=$(mktemp)
```
fmeum, line 66: Can we do this only if the arguments have a length that gets somewhat close to the limit? Otherwise we are just doing needless file I/O.

toolchain/cc_wrapper.sh.tpl
```diff
@@ -60,14 +60,19 @@ function sanitize_option() {
...
+
     while IFS= read -r opt; do
       opt="$(
         set -e
         sanitize_option "${opt}"
       )"
-      cmd+=("${opt}")
+      echo "${opt}" >> "${temp_file}"
     done <"${!i:1}"
+    mv "${temp_file}" "${!i:1}"
```
fmeum, line 75: This replaces an input file, which is something actions should avoid. Could we rewrite the argument to point to the temporary file instead?

toolchain/cc_wrapper.sh.tpl
```diff
@@ -60,14 +60,19 @@ function sanitize_option() {
...
   else
```
fmeum, line 76: The temporary file should be deleted after the compiler returns.

fmeum (commented):
Thanks for picking this up!

mering:
@fmeum I think this can be closed after #480 has been merged.

PR #432, opened 2025-01-08
Expose clang-format, git-clang-format, and libclang

AustinSchuh:
This makes it possible to use sandboxed versions of those tools in scripts.

### DIFF
--- a/toolchain/BUILD.llvm_repo
+++ b/toolchain/BUILD.llvm_repo
@@ -150,3 +150,24 @@ filegroup(
     name = "clang-tidy",
     srcs = ["bin/clang-tidy"],
 )
+
+filegroup(
+    name = "clang-format",
+    srcs = ["bin/clang-format"],
+)
+
+filegroup(
+    name = "git-clang-format",
+    srcs = ["bin/git-clang-format"],
+)
+
+filegroup(
+    name = "libclang",
+    srcs = glob(
+        [
+            "lib/libclang.so",
+            "lib/libclang.dylib",
+        ],
+        allow_empty = True,
+    ),
+)


fmeum (approved):

toolchain/BUILD.llvm_repo
```diff
@@ -150,3 +150,18 @@ filegroup(
...
+)
+
+filegroup(
+    name = "git-clang-format",
+    srcs = ["bin/git-clang-format"],
+)
+
+filegroup(
+    name = "libclang",
+    srcs = ["lib/libclang.so"],
```
fmeum, line 166: Is this always an `.so` or perhaps a `.dylib` on macOS?

AustinSchuh: You are correct.  Done now with a glob.

toolchain/BUILD.llvm_repo
```diff
@@ -150,3 +150,23 @@ filegroup(
...
+filegroup(
+    name = "git-clang-format",
+    srcs = ["bin/git-clang-format"],
+)
+
+filegroup(
+    name = "libclang",
+    srcs = glob(
+        [
+            "lib/libclang.so",
```
fmeum, line 168: This will likely fail in Bazel 8 with `--incompatible_disallow_empty_glob` because the individual globs may not match. We either need `allow_empty = True` or a wildcard.

AustinSchuh: Thanks, done.  I've been testing with bazel 7.2

fmeum (approved):

PR #433, opened 2025-01-08
Add support for extracting zst LLVM distributions

AustinSchuh:
zstd is a lot faster than xz at decompressing.  Unfortunately, llvm/clang aren't distributed with zstd, so this still takes action to create a binary package, and then add it to _llvm_distributions.

### DIFF
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -697,7 +697,12 @@ def _distribution_urls(rctx):
 
     sha256 = _llvm_distributions[basename]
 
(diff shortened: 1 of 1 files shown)


AustinSchuh:
We mirror all our packages internally with URL rewriting, and just upload a re-compressed copy to make it available.  Anyone got better ideas for that process which would work for the community?

AustinSchuh:
https://github.com/bazelbuild/bazel/issues/21210 has my benchmark numbers for size + decompression speed.  The difference is dramatic.

fmeum (approved):

PR #438, opened 2025-01-10
cc_toolchain: prefix include directories with `%workspace%`

ParkMyCar:
Today the [`built_in_includes_directores`](https://bazel.build/rules/lib/providers/CcToolchainInfo#built_in_include_directories) that are exported by this toolchain, appear to be incorrect. For example, when using `toolchains_llvm` locally with clang 18.1.8 one of the includes paths is:

```
external/llvm_toolchain/external/llvm_toolchain_llvm/include/aarch64-apple-macosx/c++/v1
```

This path does not exist, it's incorrectly prefixed with `external/llvm_toolchain/`. This is because when [populating the list](https://github.com/bazelbuild/bazel/blob/5665f76c31b8f50a56dc627b89dd3af2515c28b4/src/main/starlark/builtins_bzl/common/cc/cc_toolchain_provider_helper.bzl#L236-L238) of `built_in_includes_directories` Bazel will resolve the included directory based on [these rules](https://github.com/bazelbuild/bazel/blob/5665f76c31b8f50a56dc627b89dd3af2515c28b4/src/main/starlark/builtins_bzl/common/cc/cc_toolchain_provider_helper.bzl#L102-L122). Specifically the rule we're hitting today, and the cause of the invalid prefixing is:

> If it starts with %crosstool_top%/ or is any relative path, it is interpreted relative to the crosstool top.

This PR changes the includes paths to be prefixed with `%workspace%` so instead no prefix is added.

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -314,7 +314,7 @@ def _cc_toolchain_str(
     # visible via the "built_in_include_directories" attribute of CcToolchainInfo as well as to keep
     # them in sync with the directories included in the system module map generated for the stricter
     # "layering_check" feature.
-    toolchain_path_prefix = toolchain_info.llvm_dist_path_prefix
+    toolchain_path_prefix = "%workspace%/" + toolchain_info.llvm_dist_path_prefix
     llvm_version = toolchain_info.llvm_version
     major_llvm_version = int(llvm_version.split(".")[0])
     target_system_name = {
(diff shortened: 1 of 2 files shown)


fmeum (approved):
Nice find!

fmeum (approved):
Nice find!

ParkMyCar:
Woohoo! Thanks @fmeum!

PR #441, opened 2025-01-14
    Add target_toolchain_roots and make sysroot more hermetic

AustinSchuh:
It is useful when cross compiling to have 1 LLVM build used as the compiler, and multiple others used for their libc++/target libraries.

This concept forces us to be much more robust about what comes from the target toolchain, and what comes from the compiler.  Leave toolchain_roots for the compiler roots, but add a new flag, target_toolchain_roots with the same semantics for target toolchains. This necesitates being much more clear internally about what comes from what.

This isn't fully enough to make everything hermetic, clang still has some autodetection logic left over when using sysroots.  We want to very explicity specify the resource directory it should use, as well as where the standard C/C++ libraries come from so it finds only what we want it to find.

When pulling from a sysroot, we need to know the version of stdc++ to find includes.  Use the stdlib variable for that by supporting "stdc++-12" for example, and parsing the number out.


AustinSchuh:
https://github.com/bazelbuild/rules_cc/pull/304 needs to land and be available before this can merge.  But I would appreciate feedback on this before then as that could take a bit.

We used features before for asan/msan/cuda/etc, but I think we can use flags + select instead and avoid needing to change rules_cc again.  Feedback welcome on the best way to do all that.

brt-chong (approved):

dzbarsky:
@AustinSchuh Your goals (being very explicit about the compiler resource dir and compiler-rt vs libc++/etc) are very aligned with https://github.com/hermeticbuild/hermetic-llvm. You might want to give that one a shot if you haven't tried it yet (or let us know what's missing for you!)

AustinSchuh:
Thanks @dzbarsky.  Your goals sound similar, but I've got numerous yocto sysroots I need to cross compile for, and will need to add CUDA support for those sysroots too.  The docs on your repo suggest that is not the direction you want to go.

Long long ago (2019 or earlier), we patched the old grailbio toolchains to do all this, and never got around to upstreaming our changes.  This is one patch in a series attempting to get us back inline with upstream.

AustinSchuh:
OK, all the tests pass, and there are no merge conflicts with main.  I'd love an actual review now.

toolchain/internal/configure.bzl
```diff
@@ -159,10 +165,35 @@ def llvm_config_impl(rctx):
     use_absolute_paths_sysroot = use_absolute_paths_llvm
 
     # Check if the toolchain root is a system path.
-    system_llvm = False
-    if _is_absolute_path(toolchain_root):
+    system_llvm = _is_absolute_path(toolchain_root)
+    if system_llvm:
```
helly25: why not:
use_absolute_paths_llvm = _is_absolute_path(toolchain_root)

toolchain/internal/configure.bzl
```diff
@@ -740,6 +796,8 @@ def _is_remote(rctx, exec_os, exec_arch):
     return not (_os_from_rctx(rctx) == exec_os and _arch_from_rctx(rctx) == exec_arch)
 
 def _convenience_targets_str(rctx, use_absolute_paths, llvm_dist_rel_path, llvm_dist_label_prefix, exec_dl_ext):
+    # TODO: This doesn't really deal with cross compilation very well.  There can be multiple
+    # targets for 1 toolchain, and the target platform matters, not the exec platform.
```
helly25: So in most cases there should be a fail here or what do I get from this comment?

toolchain/rules.bzl
```diff
@@ -53,6 +53,9 @@ def llvm_toolchain(name, **kwargs):
         }
         llvm(name = name + "_llvm", **llvm_args)
 
+    if not kwargs.get("target_toolchain_roots"):
+        kwargs["target_toolchain_roots"] = kwargs.get("toolchain_roots", {})
```
helly25: Can we not do this, so we know whether it is explicitly set or unset?

toolchain/cc_toolchain_config.bzl
```diff
@@ -38,13 +38,14 @@ def cc_toolchain_config(
         target_os,
         target_system_name,
         toolchain_path_prefix,
+        target_toolchain_path_prefix,
```
helly25, line 41: We require 'target_toolchain_path_prefix' to end in '/', just like other path prefixes. But it occurred to me thta we never actually verify that, do we?

toolchain/internal/configure.bzl
AustinSchuh: I cleaned up the docs to be more clear.  This was trying to say that these are convenience targets, which make sense for builds on the target, but not cross compilation.  If someone can articulate what they should look like for cross compilation, we can fix them.

toolchain/rules.bzl
AustinSchuh: Make sense, done.  I was trying to make the default case match the target_toolchain_root case, but I found a cleaner way downstream.

toolchain/cc_toolchain_config.bzl
AustinSchuh: Verification added!  But yes, we do.  I don't think those were validated either.

AustinSchuh:
OK, looks like everything passes again now.  More changes keep getting added to main as this review stays outstanding.

I'd like to add some tests for full on cross compilation as well, but those need a couple more changes on top of this.  I figure this change is big enough as is, though I'm happy to make it whatever size makes your lives easier.

helly25 (approved):

helly25:
> OK, looks like everything passes again now. More changes keep getting added to main as this review stays outstanding.
> 
> I'd like to add some tests for full on cross compilation as well, but those need a couple more changes on top of this. I figure this change is big enough as is, though I'm happy to make it whatever size makes your lives easier.

Finally merged :-)

Yeah, we pretty much need those tests to guarantee things stay in working order. I assume a small number of tests should do it that involve Linux, Windows and MacOS. But I also expected that you were not fully ready for those -:)

PR #445, opened 2025-01-18
cc_wrapper: make built artifacts more reproducible but specifying `-ffile-prefix-map`

ParkMyCar:
This PR aims to make artifacts built with `toolchains_llvm` more reproducible by specifying:
```
-ffile-prefix-map=${{pwd}}=__bazel_toolchain_llvm_repo__
```
in the compile and linker flags. Then in `cc_wrapper.sh` we replace `${{pwd}}` with the current absolute path the script is being invoked from. This should result in builds being more reproducible by replacing absolute local paths in the binary with `__bazel_toolchain_llvm_repo__`.

For example, today in the [MaterializeInc/materialize](https://github.com/MaterializeInc/materialize/) repo, building the `@bzip2//:bzip2` target and running `strings` on the final binary, I get several results like this:
```
/private/var/tmp/_bazel_parker.timmerman/497be8b2783f71dbb03155dd9130f54f/sandbox/darwin-sandbox/977/execroot/__main__
```
After this PR all said entries are replaced with:
```
__bazel_toolchain_llvm_repo__
```

@keith I see in #440 you removed the use of `-fdebug-prefix-path`, I'm curious what you think of this since it not only effects debug info but also pre-processor macros.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -113,11 +113,12 @@ def cc_toolchain_config(
     unfiltered_compile_flags = [
         # Do not resolve our symlinked resource prefixes to real paths.
         "-no-canonical-prefixes",
-        # Reproducibility
+        # Reproducibility.
         "-Wno-builtin-macro-redefined",
         "-D__DATE__=\"redacted\"",
         "-D__TIMESTAMP__=\"redacted\"",
         "-D__TIME__=\"redacted\"",
+        "-ffile-prefix-map=${{pwd}}=__bazel_toolchain_llvm_repo__",
     ]
 
     is_xcompile = not (exec_os == target_os and exec_arch == target_arch)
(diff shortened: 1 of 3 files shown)


keith:
In general instead of using this one you should add this to your .bazelrc:

```
build --copt=-ffile-compilation-dir=. --host_copt=-ffile-compilation-dir=.
```

Ideally we would add this to this ruleset, but it wouldn't work with older versions of clang.

keith:
looks like as far back as 13.x. im not sure if anything in these rules are doing that now, but yea ideally

helly25:
> looks like as far back as 13.x. im not sure if anything in these rules are doing that now, but yea ideally

Should we just add it as soon as we detect a newer version then?

keith:
fwiw for newer clang versions `-ffile-compilation-dir=.` is preferred and requires no remapping. either way i think we should definitely do one

PR #452, opened 2025-02-05
support extra flags

rockwotj:
If you just want to add a single flag (that is per platform) and leave the rest of the default
flags, you have to copy out all the existing flags and add it to what
you have. Instead support just adding an extra flag that is appended to
the default flags.


rockwotj:
Example of where I'd like this `extra_flags` ability: https://github.com/redpanda-data/redpanda/pull/25042

toolchain/internal/repo.bzl
```diff
@@ -249,6 +249,95 @@ _compiler_configuration_attrs = {
                "target OS and arch pair you want to override " +
                "({}); empty key overrides all.".format(_target_pairs)),
     ),
+    # Same as the above flags, but instead of overriding the defaults, it just adds extras
+    "extra_compile_flags": attr.string_list_dict(
+        mandatory = False,
+        doc = ("Extra compile_flags, added after default values. " +
+               "`{toolchain_path_prefix}` in the flags will be substituted by the path " +
+               "to the root LLVM distribution directory. Provide one list for each " +
+               "target OS and arch pair you want to override " +
```
fmeum, line 258: `override` is misleading here, we are just adding flags.

fmeum (commented):
@jsharpe @rrbutani Even though this adds many attributes, they are mechanical enough that I'm okay with that. What do you think?

fmeum (approved):

rockwotj:
The alternative to all the extra flags is a way to "inject" the initial flags when providing the override flags, maybe via some magic value like $ORIGINAL_VALUES?

fmeum:
@rockwotj Sorry, I didn't notice that auto-merge didn't go through (likely because of the unresolved comment). Feel free to reopen and I can merge - I think that more attributes are better than magic values in old ones.

PR #457, opened 2025-02-10
feat: Support cross compiles from linux to macos

mrmeku:
Upstreaming an internal patch we've been using
for cross compiling the aspect CLI from a linux
machine to mac. We've been using a slightly
different version of this for quite some time
now so I cleaned up the logic significatly so it
could be upstreamed.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -168,11 +168,17 @@ def cc_toolchain_config(
     archive_flags = []
 
     # Linker flags:
-    if exec_os == "darwin" and not is_xcompile:
-        # lld is experimental for Mach-O, so we use the native ld64 linker.
-        # TODO: How do we cross-compile from Linux to Darwin?
-        use_lld = False
+    ld = "ld.lld"
+    if target_os == "darwin":
+        use_lld = True
+
+        ld = "ld64.lld"
+        ld_path = toolchain_path_prefix + "/bin/" + ld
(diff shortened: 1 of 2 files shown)


toolchain/cc_toolchain_config.bzl
```diff
@@ -168,11 +168,17 @@ def cc_toolchain_config(
...
-        # lld is experimental for Mach-O, so we use the native ld64 linker.
-        # TODO: How do we cross-compile from Linux to Darwin?
-        use_lld = False
+    ld = "ld.lld"
+    if target_os == "darwin":
+        use_lld = True
+
+        ld = "ld64.lld"
+        ld_path = toolchain_path_prefix + "/bin/" + ld
+        compile_flags.append("-mmacosx-version-min=12.0")
```
keith, line 177: this version needs to be fetched from somewhere else probably, there is a flag for this so i assume we can read it from there

fmeum (approved):

toolchain/cc_toolchain_config.bzl
```diff
@@ -228,10 +233,6 @@ def cc_toolchain_config(
         if use_lld:
             # For single-platform builds, we can statically link the bundled
             # libraries.
-            link_flags.extend([
-                "-l:libc++.a",
```
fmeum, line 232: Looks like you need to bring this back for Mac to Mac.

mrmeku: Sorry about that, that was required for our internal patch, but should not have been included in my PR

PR #459, opened 2025-02-12
Make invocation of bash more Posix conformant.

hzeller:
Hard-coding the location of the shell to be in `/bin/bash` is not always working as not all Posix-conformant operating systems (such as NixOS) provide the path /bin/bash; so shell scripts won't be able to execute.

Instead, use the `/usr/bin/env bash` idiom to reliably find the shell to execute.

This fix is only applied to the critical scripts that might be executed on the users' machine; the *_test*sh scripts are left as-is.

### DIFF
--- a/toolchain/cc_wrapper.sh.tpl
+++ b/toolchain/cc_wrapper.sh.tpl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2021 The Bazel Authors. All rights reserved.
 #
--- a/toolchain/osx_cc_wrapper.sh.tpl
+++ b/toolchain/osx_cc_wrapper.sh.tpl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2015 The Bazel Authors. All rights reserved.
 #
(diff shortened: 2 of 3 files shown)


jsharpe (approved):
LGTM

hzeller:
mmh, looks like this does not work in all environments as the CI complains.

Setting PR to draft for now and will  try to find out what is going on.

hzeller:
It looks like the `external_test` set up a very specific path that does not include a location that has bash in it (not even `/bin`). 

So in the simplest case, that `PATH` would just contain `/bin`.
But to better test that the new `/usr/bin/env` set-up is working as expected, that test would need to make bash available in some other path with only `bash` in that path and then add that to the `PATH` environment.

I have not fully understood yet where that environment (in `external_test`) with that `PATH` is set though, so will work on that as time permits (this PR is not forgotten). If someone else has some insight where that is happening (by the looks of it somewhere in `rules_rust` ?), comments welcome.

UebelAndre:
> @UebelAndre did [a fantastic job](https://github.com/bazelbuild/rules_cc/pull/466) in converting scripts in `rules_cc` to pure `/bin/sh`. How did you convert it ? Did you use tools such as `shellcheck` to figure out all the bash-isms to remove ?

I looked at what I knew to be bash-isms and looked up a pure shell way to do the same thing. I've done a similar conversion in the past to so used my experience there and then ran the tests to observe everything passing.

PR #471, opened 2025-03-12
Find LLVM distributions automatically.

helly25:
See issue: https://github.com/bazel-contrib/toolchains_llvm/issues/473

Instead of implementing a complex way of predicting the basename find them by common prefixes in the database we anyway have.

Rather then forcing maintainers to enter more complex information this keeps us at simply adding the versions as generated from the hasher script.

This should also ultimately put issue https://github.com/bazel-contrib/toolchains_llvm/issues/464 and any future occurrence to rest.

The results of the new search based algorithm are generally be better. The algorithm removes a lot of the mistakes from the old implementation, is more correct in some cases and performs other new correct decisions. Yet, the new algorithm may not always be perfect, as it may have inherited some bad decisions of the old algorithm.

As the diff in `toolchain/internal/llvm_distributions.golden.out.txt` shows we only drop `del:` lines, which means we can find more distributions.

The diff in `` is harder to interpret. The diffs are best viewed side by side as both old and new file are supposed to have the same contents. At least their "keys" are in the same order. I use a custom diff program that is designed for this. It strips the actual error messages. When only looking at new errors (as shown below), then we see no difference:

```sh
bazel run //mbo/diff -- --regex_replace_lhs='/(.*)(ERROR).*/\1\2/' --regex_replace_rhs='/(.*)(ERROR).*/\1\2/' ~/llvm_distributions.golden.sel.orig.txt ~/llvm_distributions.golden.sel.find.txt --algorithm=direct | awk '{l=c;c=$0}/^+.*ERROR/{print l"\n"c}'
```

If we look at all differences, then we see some 782 former errors now resulting in a distribution. They now find the likely best choice.

Some of these can be improved. But imo it is not worth doing so.

The PR does not change anything regarding manually provided information. But some folks will be able to switch to automatic selection instead.


helly25:
@fmeum @jsharpe @rrbutani

Hi, what is the best way to discuss the following:
* https://github.com/bazel-contrib/toolchains_llvm/issues/473
* https://github.com/bazel-contrib/toolchains_llvm/pull/472
* this PR

fmeum:
Could you open a discussion and share it in the #cc channel? I don't think I'm a heavy enough user to decide whether this is useful or not.

helly25:
> Could you open a discussion and share it in the #cc channel? I don't think I'm a heavy enough user to decide whether this is useful or not.

See https://github.com/bazel-contrib/toolchains_llvm/discussions/485

toolchain/internal/llvm_distributions.bzl
```diff
@@ -741,19 +736,268 @@ def _get_llvm_version(rctx):
...
 def _find_llvm_basename_list(llvm_version, host_info):
     """Lookup (llvm_version, arch, os) in the list of basenames in `_llvm_distributions.`"""
-    name, _ = llvm_release_name_host_info(llvm_version, host_info)
-    if name in _llvm_distributions:
-        return [name]
+    arch = host_info.arch
+    os = host_info.os
+    dist = host_info.dist
+
+    # Prefer new LLVM- distributions is available
```
fmeum, line 857:
```suggestion
    # Prefer new LLVM distributions if available
```

fmeum (approved):

helly25:
@fmeum thanks for the approval. I committed your suggestion. Once this is merged I will attempt to simplify things a bit and maybe add or remove a few choices as appropriate.

PR #472, opened 2025-03-15
Implement latest requirements

helly25:
See issue https://github.com/bazel-contrib/toolchains_llvm/issues/473

This builds on PR https://github.com/bazel-contrib/toolchains_llvm/pull/471

This PR allows requirements support for the LLVM versions like `llvm_version = "first"` and `llvm_version = "latest"` which will effectively find the first and latest LLVM version respectively. For many people this will be the easiest way to get started and solves the issue that not always all arch/os versions have been prepared for all LLVM versions. It also means you do not have to know whether the basename starts with "LLVM" or "clang_llvm", in which order arch and os have to be provided and how the os is exactly specified.

We further add the ability to specify requirements, including minimum and maximum versions as well as version exclusions. That can be easier than figuring out which versions are available for which platform. Further this allows to automatically pick up relevant versions automatically as they become available later. Example:

```
llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
llvm.toolchain(
    name = "llvm_toolchain_llvm",
    llvm_version = "latest:>=19.1.6,!=19.1.7,!=20.1.0",
)
use_repo(llvm, "llvm_toolchain_llvm")
```

The requirements strings mostly follow semver specs and support `<`, `<=`, `>`,  `>=`, `==` and `!=`. They only currently work for version components `major`, `minor` and `patch`. Since we do not list other versions in our config that is fine for now.

One practical goal for this PR is to be able to run automatic tests with updated LLVM versions in the easiest possible way. Using the new extra_llvm_versions we can now easily run automated CI/CD tests against different LLVM branches. Say:
* `llvm_version = 'latest:>=15,<16'`
* `llvm_version = 'latest:>=19,<20'`
* `llvm_version = 'latest:>=20,<21'`
That above list would be used in three separate CI/CD runs that would then test the 15, 19 and 20 branches respectively. When new distributions get added upstream or locally, they can easily be added via the new `extra_llvm_distributions` and will be picked up by the setup without any further change.


helly25:
The current error with workspaces mode is in the external dependency, so a new version is required to fix that -> helly25_bzl >= 0.1.2.

helly25:
@fmeum this is ready from my end now, I'd appreciate any feedback. I added a concrete practical goal to the PR description.

helly25:
> This looks good to me after a quick read through. I'm also fine with the new dep that handles version parsing.
> 
> Could you resolve the conflict and perform the cleanup you had in mind? Then I can do a full review.

@fmeum All merged today. Trying to get a green run. Looks like MacOS runners have an allocation problem. Maybe reruns fix that.

PR #477, opened 2025-04-01
Add initial support for linux-armv7

UebelAndre:
This change allows users to target `linux-armv7` without running into an explicit fail due to missing definitions.


UebelAndre:
This could replace https://github.com/bazel-contrib/toolchains_llvm/pull/437

PR #480, opened 2025-04-09
Create sanitized response file instead of expanding arguments

### LINKED ISSUE #421: Why are response files expanded in `osx_cc_wrapper`?
I'm trying to build a relatively large Rust project with `rules_rust` and LTO enabled on macOS, but the linking step fails with:
```
= note: external/llvm_toolchain/bin/cc_wrapper.sh: line 129: external/llvm_toolchain_llvm/bin/clang: Argument list too long
```

I enabled logging for `rustc` and I can see that it spills arguments to a response file, i.e.

```
INFO rustc_codegen_ssa::back::link falling back to passing arguments to linker via an @-file
```

But then it looks like `osx_cc_wrapper` (and FWIW the Linux version too) [expands response files](https://github.com/bazel-contrib/toolchains_llvm/blob/master/toolchain/osx_cc_wrapper.sh.tpl#L89) which ultimately leads to the "Argument list too long error". With how long Bazel files paths can be, and the way Rust parallelizes compilation with codegen units, the argument list can get long very quickly.

Is it possible to stop expanding response files? Or is there some reason that the wrapper needs to? Thanks in advance!

dfreese:
The cc_wrapper.sh scripts inspects response files and sanitizes each line within them.  How the logic is structured currently causes all of these arguments to the be directly used later in the script.  This can run into argument overflows.  It can also run into problems where quoting in response files and from args is treated differently.

This addresses some of the comments in #430, namely cleaning up files and not overwriting the input files.  I have not tried to apply a threshold here as that is best left up to the tooling that previously decided to use response files, which can cause difficult to debug changes in behavior.

This conflicts with #479, but is trying to address the same underlying problem of quoted arguments in response files as the result of golang's link actions.

Fixes #421

### DIFF
--- a/toolchain/cc_wrapper.sh.tpl
+++ b/toolchain/cc_wrapper.sh.tpl
@@ -29,6 +29,16 @@
 
 set -euo pipefail
 
+CLEANUP_FILES=()
+
+function cleanup() {
+  if [[ ${#CLEANUP_FILES[@]} -gt 0 ]]; then
+    rm -f "${CLEANUP_FILES[@]}"
+  fi
+}
+
+trap cleanup EXIT
+
 # See note in toolchain/internal/configure.bzl where we define
 # `wrapper_bin_prefix` for why this wrapper is needed.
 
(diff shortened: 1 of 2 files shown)


dfreese:
added a commit to address formatting errors and the unused variable.  For me that isn't showing up on the PR due to github "Processing Updates"

dfreese:
Force-pushed to get retry the "Processing updates".  Should be good to go now.

mering:
> This conflicts with #479, but is trying to address the same underlying problem of quoted arguments in response files as the result of golang's link actions.

I think we still want both (this to avoid too long argument lines and the other one to correctly sanitize options within a response file).

fmeum:
@dfreese Please fix the remaining linter findings, then this should be good to go in.

fmeum (approved):

ParkMyCar:
Thanks @dfreese!

PR #483, opened 2025-04-17
Probe each toolchain bin for existence

scasagrande:
Probe each toolchain binary for existence before attempting to make symlinks to them.

This allows for llvm packages without tools that aren't needed by a given project.

A simple loop here was used for readability with consideration for the limited number of items in the loop.

Here is an example copy of `tools` after the filtering when using this package: https://github.com/MaterializeInc/toolchains/releases/tag/clang-18.1.8-5

```
 {"clang-cpp": "clang-cpp", "ld.lld": "ld.lld", "llvm-ar": "llvm-ar", "llvm-dwp": "llvm-dwp", "llvm-profdata": "llvm-profdata", "llvm-cov": "llvm-cov", "llvm-nm": "llvm-nm", "llvm-objcopy": "llvm-objcopy", "llvm-objdump": "llvm-objdump", "llvm-strip": "llvm-strip"}
```

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -121,7 +121,12 @@ def llvm_config_impl(rctx):
         # symlinked path from the wrapper.
         wrapper_bin_prefix = "bin/"
         tools_path_prefix = "bin/"
-        tools = _toolchain_tools(os)
+        # Probe each binary for existence
+        tools = {}
+        for tool in _toolchain_tools(os).items():
+            if rctx.path(llvm_dist_rel_path + "bin/" + tool[0]).exists:
+                tools.update([tool])
+
         for tool_name, symlink_name in tools.items():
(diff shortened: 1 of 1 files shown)


scasagrande:
Looks like all the failing tests are upset about llvm-ar

```/home/runner/.cache/bazel/_bazel_runner/f764bf27ef94a0d5a3cfad3bad3e40be/sandbox/processwrapper-sandbox/526/execroot/_main/external/toolchains_llvm++llvm+llvm_toolchain/bin/llvm-ar: not found```

I'll keep looking

scasagrande:
Oh, it was working locally for me because the download and extraction had already completed.

Running the test, then making a trivial change to `toolchain/internal/configure.bzl` and running the test again results in a pass. This code is executed before we download and extract the package. I'll take another look later if no one else has any suggestions.

fmeum:
If it helps, we could always create the symlinks and then write out a `.bzl` with the tools that actually exist after the download has happened.

PR #486, opened 2025-04-25
Add missing distributions and check for omissions

helly25:
Add missing distributions.
Starting with 19.0.0 verify which distributions we can actually automatically detect.


toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +757,101 @@ def _distribution_urls(rctx):
...
+    out = ctx.actions.declare_file(ctx.label.name + ".out")
+    ctx.actions.write(out, "\n".join(output) + "\n")
+    return [
+        DefaultInfo(files = depset([out])),
+        DistrubutionsInfo(out = out),
+    ]
+
+write_distributions = rule(
+    implementation = _write_distributions_impl,
+    output_to_genfiles = True,
```
fmeum, line 855:
```suggestion
```
This is deprecated and should not be used.

toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +757,101 @@ def _distribution_urls(rctx):
...
+    provides = [DefaultInfo, DistrubutionsInfo],
```
fmeum, line 856: This is only necessary for aspects, which we don't have a need for here. 

```suggestion
```

toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +757,101 @@ def _distribution_urls(rctx):
     strip_prefix = strip_prefix.rstrip("-rhel86")
 
     return urls, sha256, strip_prefix
+
+DistrubutionsInfo = provider(
```
fmeum, line 761: Do we need this? It looks like it's essentially unused and `DefaultInfo` is enough.

toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +757,101 @@ def _distribution_urls(rctx):
...
+    },
+)
+
+def _version_ge(lhs, rhs):
+    _lhs = lhs.split(".")
+    _rhs = rhs.split(".")
+    for p in range(3):
+        if int(_lhs[p]) < int(_rhs[p]):
+            return False
+    return len(_lhs) >= len(_rhs)
```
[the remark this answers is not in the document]
helly25: Good idea :-)

toolchain/internal/llvm_distributions.bzl
helly25: It was an experimentation left over. And I don't think we need to do anything with aspects or more complex than the test we have. So all gone.

toolchain/internal/llvm_distributions.bzl
helly25: Done.

toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +758,87 @@ def _distribution_urls(rctx):
...
+            name = name.removeprefix(prefix)
+        version = name.split("-", 1)[0]
+        if not _version_ge(version, MIN_VERSION):
+            continue
+        version_list.append(version)
+    for version in _llvm_distributions_base_url.keys():
+        if not _version_ge(version, MIN_VERSION):
+            continue
+        version_list.append(version)
+    versions = {v: v for v in version_list}
```
fmeum, line 801:
```suggestion
    versions = {v: None for v in version_list}
```

toolchain/internal/llvm_distributions.bzl
```diff
@@ -714,3 +758,87 @@ def _distribution_urls(rctx):
...
+
+    # Write versions to output to check which versions we take into account.
+    output = []
+    for version in versions:
```
fmeum, line 805:
```suggestion
    for version in versions.keys():
```

fmeum (approved):
Thanks for the great attention to detail. This will make maintenance much easier.

PR #488, opened 2025-04-26
Windows exe

helly25:
For V19+:
* add win32.exe, win64.exe and woa.exe distributions.
* find win32.exe and woa64.exe distributions automatically.

The exe files are self extracting installers, not msiexe. However, we can extract them with a separate tool, say 7z.


fmeum:
The change looks good, but is toolchains_llvm set up to do anything with an `.exe`? I've never used LLVM on Windows, but isn't that an installer?

helly25:
> The change looks good, but is toolchains_llvm set up to do anything with an `.exe`? I've never used LLVM on Windows, but isn't that an installer?

I'm still at my mac/linux machines and will have to figure out how I can actually teach the toolchain to work with them. Unfortunately Bazel does not use a tool for such SFX files and the toolchains will need to add support in the download implementation.

At the moment our `_windows` functions will return either `LLVM-{llvm_version}-win32.exe` or ``LLVM-{llvm_version}-win64.exe` which is incorrect and wrong in several ways. The PR in its current form at least is a step forward.

But I'll need to work upstairs later with a windows machine to see how I can get it actually working.

PR #489, opened 2025-04-26
Tweaks

helly25:
Tweaks:
* Sort known distro names.
* Function `_os_version_arch` returns a dist(ribution) name and not an os name, so we rename it `_dist_version_arch`.
* Move `_resolve_version_for_suse` above `_linux` to follow order of the other helpers.
* Improve strip_prefix detection (also added `.exe` as we predict it, though we cannot really handle it (yet)).

### DIFF
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -748,11 +748,12 @@ def _distribution_urls(rctx):
 
     sha256 = _llvm_distributions[basename]
 
-    if basename.endswith(".tar.xz"):
-        strip_prefix = basename[:(len(basename) - len(".tar.xz"))]
-    elif basename.endswith(".tar.zst"):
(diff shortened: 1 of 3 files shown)


toolchain/internal/llvm_distributions.bzl
```diff
@@ -748,11 +748,12 @@ def _distribution_urls(rctx):
...
 
-    if basename.endswith(".tar.xz"):
-        strip_prefix = basename[:(len(basename) - len(".tar.xz"))]
-    elif basename.endswith(".tar.zst"):
-        strip_prefix = basename[:(len(basename) - len(".tar.zst"))]
-    else:
+    strip_prefix = ""
+    for suffix in [".exe", ".tar.gz", ".tar.xz", ".tar.zst"]:
+        if basename.endswith(suffix):
+            strip_prefix = basename.rstrip(suffix)
```
fmeum, line 754:
```suggestion
            strip_prefix = basename.removesuffix(suffix)
```

PR #490, opened 2025-04-27
Verify the current resolution

helly25:
* Split `llvm_release_name` into two:
  * `llvm_release_name_host_info`: Used to verify the resolution. It does not take a rctx.
  * `llvm_release_name_context`: Will fail if resolution from rctx fails.
* Change `dist_version_arch` to `host_info` which returns a struct, so the result can easily be extended and only one call is needed.

As can be seen in `toolchain/internal/llvm_distributions.golden.sel.txt` there are a number of decisions taken in mode `auto` that lead to failures in predicting a correct LLVM distribution. Those we should fix.


helly25:
@fmeum Any chance you can review and we can progress on this? I want to use this as a basis and verification of improvements to finding targets automatically.

toolchain/internal/BUILD.bazel
```diff
@@ -20,11 +20,34 @@ exports_files(["template.modulemap"])
...
+    name = "llvm_distributions_select_no_error_test",
+    srcs = ["llvm_distributions_select_no_error_test.sh"],
+    data = [
+        "llvm_distributions.golden.sel.txt",
+        "llvm_distributions.sel.txt",
+    ],
+    # Restrict to Linux. While this would also work for MacOS the restriction
+    # only works for one platform. To select two platforms we would need to use
+    # `select`. That can be made work with `tag = select...manual`.
+    target_compatible_with = [
```
fmeum, line 50: I mostly work on macOS and would like test coverage to be as good as possible on that platform. Could you use `select` and `@platforms//: incompatible` to make this work on both macOS and Linux?

toolchain/internal/common.bzl
```diff
@@ -103,15 +103,23 @@ def _linux_dist(rctx):
...
+def host_info(rctx):
     _os = os(rctx)
     _arch = arch(rctx)
 
     if _os == "linux" and not rctx.attr.exec_os:
-        (distname, version) = _linux_dist(rctx)
-        return distname, version, _arch
-
-    return _os, "", _arch
+        (dist_name, dist_version) = _linux_dist(rctx)
```
fmeum, line 111:
```suggestion
        dist_name, dist_version = _linux_dist(rctx)
```

toolchain/internal/llvm_distributions.bzl
```diff
@@ -785,25 +807,74 @@ def _write_distributions_impl(ctx):
...
+        "x86_32",
         "x86_64",
     ]
     os_list = [
         "darwin",
         "linux",
-        "raspbian",
         "windows",
     ]
+    ANY_VER = "0"  # Version does not matter, but must be valie integer
```
fmeum, line 824:
```suggestion
    ANY_VER = "0"  # Version does not matter, but must be a valid integer
```

toolchain/internal/llvm_distributions.bzl
```diff
@@ -785,25 +807,74 @@ def _write_distributions_impl(ctx):
...
+    MIN_VERSION = "6.0.0"
     version_list = []
     for name in _llvm_distributions.keys():
         for prefix in ["LLVM-", "clang+llvm-"]:
             if name.startswith(prefix):
                 version = name.split("-", 2)[1]
-                if not _version_ge(version, MIN_VERSION):
-                    continue
-                version_list.append(version)
+                if _version_ge(version, MIN_VERSION):
```
fmeum, line 875: Could we parse the versions once and then compare with `>=`? That should be faster and also more readable.

toolchain/internal/llvm_distributions.bzl
```diff
@@ -827,28 +899,74 @@ def _write_distributions_impl(ctx):
...
-                basenames = _find_llvm_basename_list(version, arch, os)
-                if len(basenames) != 1:
-                    continue
-                basename = basenames[0]
-                if basename in _llvm_distributions:
-                    if basename in not_found:
-                        not_found.pop(basename)
-                else:
-                    result[basename] = True
+                dist_list = dist_dict_list.get(os, [struct(name = os, version = "")])
```
fmeum, line 909: How does the empty string differ from ANY_VER?

toolchain/internal/llvm_distributions.bzl
fmeum, line 824:
```suggestion
    ANY_VERSION = "0"  # Version does not matter, but must be valie integer
```

toolchain/internal/llvm_distributions.bzl
```diff
@@ -13,8 +13,18 @@
 # limitations under the License.
 
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc", "use_netrc")
-load("//toolchain/internal:common.bzl", _arch = "arch", _attr_dict = "attr_dict", _exec_os_arch_dict_value = "exec_os_arch_dict_value", _os = "os")
-load("//toolchain/internal:release_name.bzl", _llvm_release_name = "llvm_release_name", _llvm_release_name_19 = "llvm_release_name_19")
+load(
+    "//toolchain/internal:common.bzl",
+    _attr_dict = "attr_dict",
```
fmeum, line 18: You can drop all the underscores and renames, loads aren't reexported automatically.

toolchain/internal/release_name.bzl
```diff
@@ -111,11 +106,14 @@ def _ubuntu_osname(arch, version, major_llvm_version, llvm_version):
         elif llvm_version in ["7.1.0"]:
             os_name = "linux-gnu-ubuntu-14.04"
         else:
-            fail("LLVM patch release %s not available for Ubuntu %s" % (llvm_version, version))
+            error = "ERROR: LLVM patch release %s not available for Ubuntu %s" % (llvm_version, version)
+            if should_fail:
```
fmeum, line 110: Could we refactor this method to use Go-style `return value, err`? Then the caller can decide whether to fail and we don't have to cram errors into a string.

toolchain/internal/llvm_distributions.golden.out.txt
```diff
@@ -0,0 +1,496 @@
+version: 6.0.0
```
fmeum, line 1: This file looks fine and I see why it makes sense to have, but could you explain why we need the other (massive) one?

helly25: The larger one is only necessary until we are happy with the selection. Once that is done we can drop the file and related test.

toolchain/internal/llvm_distributions.bzl
helly25: Above an actual integer version is required. Here we just need a value.
Imo it is better to leave the above 0 and empty here so we know whether something was provided or is a default inserted due to failed lookup.
It might be possible to change, but then all golden files change for no reason.

toolchain/internal/llvm_distributions.bzl
helly25: Good idea, I did not like it, but kept it for consistency.

toolchain/internal/BUILD.bazel
helly25: Done

toolchain/internal/llvm_distributions.bzl
helly25: Done.

toolchain/internal/llvm_distributions.bzl
helly25: Done.

toolchain/internal/release_name.bzl
helly25: Done.

toolchain/internal/llvm_distributions.golden.out.txt
fmeum: On mobile right now, but if that file really has 20k lines, how are we going to review it? What would I be looking for?

helly25: Ah. The file itself is just a snapshot of our current decisions without the need to actually run on each machine. The file itself allows to verify two things:
1) Check whether a line that contains an ERROR should probably produce a valid result, and
2) You can manually look for decisions that are wrong.
More importantly though, the file must be changed before any change to the select algorithm. At that point we see exactly which decisions change and how. So most likely we will only look at differences. Further, there is an already filtered test that only contains non error decisions to simplify review.

helly25:
@fmeum any other comments/questions/concerns?

fmeum (approved):

PR #495, opened 2025-05-11
Add extra_llvm_distributions attribute

helly25:
This simplifies adding new versions without upgrading the toolchain.

Tested:
* Removed 20.1.4 and added those instead to the config of a repo using the toolchain.
* Added the config for 20.1.4 back to see whether the config can be present in the toolchain and in the user repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,25 @@ We currently offer limited customizability through attributes of the
 [llvm_toolchain\_\* rules](toolchain/rules.bzl). You can send us a PR to add
 more configuration attributes.
 
-The following shows how to add a specific version for a specific target before
-the version was added to [llvm_distributions.bzl](toolchain/internal/llvm_distributions.bzl):
+The `MODULE.bazel` example below demonstrates how to add new LLVM distributions before the toolchain has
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,8 +50,25 @@ We currently offer limited customizability through attributes of the
...
-the version was added to [llvm_distributions.bzl](toolchain/internal/llvm_distributions.bzl):
+The `MODULE.bazel` example below demonstrates how to add new LLVM distributions before the toolchain has
+been updated. They can easily be computed using the provided checksum tool (see `llvm_checksums.sh -h`).
+
+```starlark
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
+llvm.toolchain(
+    name = "llvm_toolchain",
+    llvm_version = "20.1.4",
+    llvm_distributions = {
```
fmeum, line 61: What do you think of calling this `extra_llvm_distributions` to emphasize that the default distributions may still be used?

fmeum (approved):

README.md
helly25: Ah, good point! My name kind of indicates a replacement :-)

PR #501, opened 2025-05-17
Allow urls in llvm_extra_distributions

helly25:
In extra_llvm_distributions allow URLs and absolute filenames. If absolute filenames are used, then we turn them into "file:" urls which are also understood.

If strip_prefix is not given, then we compute it.

Extend tests to verify llvm_extra_distributions actually work.


fmeum (approved):

helly25:
> Sorry for the delay, it's a busy period for me

NW. I saw you're working on some super-duper cool stuff :-)

PR #502, opened 2025-05-21
Add support for x86_64-unknown-none targets

jblebrun:
By default, no sysroot is used.

This is useful for bare-metal targets, firmwares, etc.

We are actively using this patch in https://github.com/project-oak/oak

### DIFF
--- a/platforms/BUILD.bazel
+++ b/platforms/BUILD.bazel
@@ -23,6 +23,14 @@ platform(
 )
 
 platform(
+    name = "none-x86_64",
+    constraint_values = [
+        "@platforms//os:none",
+        "@platforms//cpu:x86_64",
+    ],
+)
+
+platform(
     name = "linux-aarch64",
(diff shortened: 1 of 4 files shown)


fmeum (approved):

fmeum:
Could you look into the CI failures?

jblebrun:
> I can't reproduce the failure locally, so I created commenting this out and pushing again.

Update: I found a local repro. I will figure out what's happening and push a new update. So you can hold off on enabling the current CI run.

jblebrun:
Ok, I figured it out. Checking on `os == "none"` alone was also allowing wasm32|64-none-none to pass.

To fix it, I've created a list of target pairs that should support no-sysroot.

It seems it should be theoretically possible to support wasm-none as well, but that might be a deeper change?

jblebrun:
Thanks!

PR #530, opened 2025-08-05
Add support for riscv32-unknown-none-elf targets

kitterion:
This adds support for bare metal riscv32 targets that don't need a sysroot.

I have an esp32 riscv board that I'm building for, and to set up an appropriate toolchain, `toolchains_llvm` needs to know about it.

### DIFF
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -306,7 +306,7 @@ def _cc_toolchain_str(
         if exec_os == target_os and exec_arch == target_arch:
             # For darwin -> darwin, we can use the macOS SDK path.
             sysroot_path = _default_sysroot_path(rctx, exec_os)
-        elif target_pair in _supported_no_sysroot_targets:
+        elif (target_os, target_arch) in _supported_no_sysroot_targets:
             sysroot_path = ""
         else:
             # We are trying to cross-compile without a sysroot, let's bail.
@@ -331,6 +331,7 @@ def _cc_toolchain_str(
         "linux-aarch64": "aarch64-unknown-linux-gnu",
         "linux-armv7": "armv7-unknown-linux-gnueabihf",
         "linux-x86_64": "x86_64-unknown-linux-gnu",
+        "none-riscv32": "riscv32-unknown-none-elf",
         "none-x86_64": "x86_64-unknown-none",
         "wasm32": "wasm32-unknown-unknown",
         "wasm64": "wasm64-unknown-unknown",
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -97,6 +97,14 @@ def cc_toolchain_config(
             "clang",
             "glibc_unknown",
         ),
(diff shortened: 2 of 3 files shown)


toolchain/internal/configure.bzl
```diff
@@ -306,7 +306,7 @@ def _cc_toolchain_str(
         if exec_os == target_os and exec_arch == target_arch:
             # For darwin -> darwin, we can use the macOS SDK path.
             sysroot_path = _default_sysroot_path(rctx, exec_os)
-        elif target_pair in _supported_no_sysroot_targets:
```
kitterion, line 309: I'm pretty sure this never worked. `target_pair` is a string (`"none-riscv32"` in my case) but `_supported_no_sysroot_targets` is a list of pairs of strings.

kitterion:
That sounds great. The issue is currently it's a bit awkward to specify the particular flavour of riscv one wants to use. For example, my riscv32 toolchain declaration looks like this

```bzl
llvm.toolchain(
    name = "llvm_toolchain_esp32_c6",
    llvm_version = LLVM_VERSION,
    stdlib = {"none-riscv32": "none"},
    compile_flags = {
        "none-riscv32": [
            # https://github.com/espressif/esp-idf/blob/v5.5/tools/cmake/toolchain-clang-esp32c6.cmake#L14
            "-march=rv32imac_zicsr_zifencei",
            "-mabi=ilp32",

            # Default flags
            "--target=riscv32-unknown-none-elf",
            "-U_FORTIFY_SOURCE",
            "-fstack-protector",
            "-fno-omit-frame-pointer",
            "-fcolor-diagnostics",
            "-Wall",
            "-Wthread-safety",
            "-Wself-assign",
        ],
    },
    link_flags = {
        "none-riscv32": [
            # https://github.com/espressif/esp-idf/blob/v5.5/tools/cmake/toolchain-clang-esp32c6.cmake#L14
            "-march=rv32imac_zicsr_zifencei",
            "-mabi=ilp32",

            # Default flags
            "--target=riscv32-unknown-none-elf",
            "-no-canonical-prefixes",
            "-fuse-ld=lld",
            "-Wl,--build-id=md5",
            "-Wl,--hash-style=gnu",
            "-Wl,-z,relro,-z,now",
            "-nostdlib",
        ],
    },
)
llvm.sysroot(
    name = "llvm_toolchain_esp32_c6",
    targets = ["none-riscv32"],
    label = "//:empty",
)
llvm.extra_target_compatible_with(
    name = "llvm_toolchain_esp32_c6",
    targets = ["none-riscv32"],
    constraints = ["//platforms:esp32_c6_constraint"],
)
```

A few notes about that:
- I might have missed an option, but since `compile_flags` and `link_flags` override what's already there, I had to explicitly specify the flags that I'm overriding
- label `//:empty` does not exist in the root BUILD.bazel file. I just had to add an `llvm.sysroot` declaration for toolchains_llvm to be happy

I'm not sure `toolchains_llvm` wants to advertise all of that. It might be better to fix the rough edges first, and then add the documentation.

PR #531, opened 2025-08-06
Mark downloaded repository as reproducible

trevorgray:
#515 only makes the config repository reproducible. The downloaded toolchains are store in a separate repository. Since this separate repository is not marked as reproducible, the extracted files are currently not cached.

This change fixes that by marking the repositority that contains the downloaded toolchain also reproducible.

Similar to the git repo rule here: https://github.com/bazelbuild/bazel/blob/648f991bd8653408ba6b4c24c6ff8fa4fce51eba/tools/build_defs/repo/git.bzl#L219


fmeum (approved):

fmeum (approved):

fmeum (approved):
Nice, thanks for digging!

fmeum (approved):

PR #539, opened 2025-08-27
Add LLVM release 21.1.0.

helly25:
https://github.com/bazel-contrib/toolchains_llvm/issues/538


helly25:
This needs either a switch from debian:latest to debian:bookworm or a fix to allow either option. For now the new version is imo more relevant and we should go with https://github.com/bazel-contrib/toolchains_llvm/pull/541 to enable this PR. After that we should possibly have separate tests for debian:bookworm and debian:trixie.

PR #543, opened 2025-09-01
replace `/bin/bash` shebang with `/usr/bin/env bash`

oliverlee:
This allows systems which don't have this location (e.g. NixOS) to run the scripts.

### DIFF
--- a/tests/file_dependency_test.sh
+++ b/tests/file_dependency_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright 2022 The Bazel Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/tests/scripts/archlinux_test.sh
+++ b/tests/scripts/archlinux_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright 2018 The Bazel Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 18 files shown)


oliverlee:
I suppose the failures are related to this:
https://github.com/bazel-contrib/toolchains_llvm/blob/0bdeae25025aeb6f914db2a02cb2efeb302743b5/toolchain/osx_cc_wrapper.sh.tpl#L165-L167

RyanTorok:
As you mostly pointed out already, using `/usr/bin/env bash` is a double-edged sword. On one hand, it's guaranteed to exist on all [POSIX-compliant ](https://discourse.nixos.org/t/add-bin-bash-to-avoid-unnecessary-pain/5673/2) systems, so it will be present on, e.g. NixOS. On the other hand, it's entirely reliant on what the calling ruleset puts in `$PATH`.

I'm not too familiar with the internals of `rules_rust` either, but based solely on the comment you tagged above, it appears that `rules_rust` doesn't add `/usr/bin` to `$PATH`, which means `/usr/bin/env` won't find `/usr/bin/bash`, even though the latter exists.

I'm not sure if this is the cleanest solution possible, but if you really need `bash` for these scripts, you might consider having a small wrapper script that uses `!#/bin/sh` (which is required to exist on POSIX-compliant systems) and searches manually for the right shell. You could default to `/bin/bash` and then fall back to `/usr/bin/env bash` if the former doesn't exist. Then just invoke the existing script using the shell you find.

Hope this helps!

oliverlee:
Thanks for your comments.

> It's really only the two cc_wrapper scripts that matter.

Probably only `toolchain/cc_wrapper.sh.tpl` matters. If `toolchain/osx_cc_wrapper.sh.tpl` is only used on macOS, we can probably assume `/bin/bash` exists.

I'm not sure I'll have time to look into this in the near term, so feel free to take over if you're interested.

PR #568, opened 2025-10-08
Don't forbid use_extension("llvm.bzl") inside non-root modules

EdSchouten:
Though I understand why the authors of this module don't want to permit module extension tag declarations of llvm.toolchain() in non-root modules, not being able to depend on the extension entirely is fairly restrictive.

In one of my Bazel projects I want to be able to use clang-format as part of some reformatting script. These scripts are not just used by the module itself, but also by some other modules that depend on it. I therefore have the following stanza in MODULE.bazel:

    llvm_dev = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency = True)
    llvm_dev.toolchain(
        llvm_version = "17.0.6",
    )
    llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
    use_repo(llvm, "llvm_toolchain_llvm")

I can then use label @llvm_toolchain_llvm//:bin/clang-format to get a copy of clang-format. The current logic prevents me from doing this, because the second use_extension() call is rejected outright, even if it doesn't invoke llvm.toolchain().

### DIFF
--- a/toolchain/extensions/llvm.bzl
+++ b/toolchain/extensions/llvm.bzl
@@ -50,10 +50,10 @@ def _constraint_dict(tags, name):
 
 def _llvm_impl_(module_ctx):
     for mod in module_ctx.modules:
-        if not mod.is_root:
-            fail("Only the root module can use the 'llvm' extension")
         toolchain_names = []
         for toolchain_attr in mod.tags.toolchain:
+            if not mod.is_root:
+                fail("Only the root module can use the 'llvm.toolchain()' tag")
             name = toolchain_attr.name
             toolchain_names.append(name)
             attrs = {


fmeum (commented):
I am slightly worried about the availability of binary tools changing with LLVM versions.

In your module that uses clang-format, is that reformatting script something you would expect to run when that module isn't the root module? If it's a dev tool, wouldn't it be fine to mark both extension usages as dev deps?

If that's not possible for some reason, I'm open to merging this.

helly25:
> It’s a reformatting script that I keep in bb-storage that’s also capable of reformatting code in any other bb-* repo.
> 
> So I’m not able to mark it dev_dependency=True

I assume you are doing something like the following:

```
"_clang_format_tool": attr.label(
        doc = "The target of the clang-format executable.",
        default = Label("@llvm_toolchain_llvm//:bin/clang-format"),
        allow_single_file = True,
        executable = True,
        cfg = "exec",
    ),
```

If you are doing something different, mind sharing?

PR #572, opened 2025-10-21
Add a repository rule for efficient sysroots

dzbarsky:
This provides sysroots as a source directory rather than a filegroup, which means merkle tree computation will be more efficient

### DIFF
--- a/toolchain/sysroot.bzl
+++ b/toolchain/sysroot.bzl
@@ -12,19 +12,11 @@ def _sysroot_impl(rctx):
 
     rctx.download(urls, archive, sha256 = rctx.attr.sha256)
 
-    # Sysroot handling has assumptions about the filegroup's package matching the sysroot directory,
-    # but provide an alias to handle the existing usage of the `//:sysroot` target.
-    rctx.file(
-        "BUILD.bazel",
-        """alias(
-    name = "sysroot",
-    actual = "//sysroot",
-    visibility = ["//visibility:public"],
-)""",
-    )
+    # Source directories are more efficient than file groups for 2 reasons:
+    #   - They can be symlinked into a local sandbox with a single symlink instead of 1-per-file
+    #   - They serve as a signal to the Merkle tree cache machinery since they can be memoized as a single node.
+    # Since sysroots are usually a ton of files, it can improve build performance to declare them as source directories.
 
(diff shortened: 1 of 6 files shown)


toolchain/sysroot.bzl
```diff
@@ -0,0 +1,50 @@
...
+    _, _, archive = url.rpartition("/")
+
+    rctx.download(url, archive, sha256 = rctx.attr.sha256)
+
+    # Declare the sysroot files as a source directory so they can be
+    # optimized in the Merkle tree cache more effectively.
+    rctx.file("sysroot/BUILD.bazel", \
+"""filegroup(
+    name = "sysroot",
+    srcs = ["."],
```
fmeum: Super nit: this will also stage the BUILD file. Could we extract into a subdirectory instead?

fmeum: So sneaky :⁠-⁠) could you add this as a comment?

MODULE.bazel
```diff
@@ -21,9 +21,19 @@ module(
...
+bazel_dep(name = "aspect_bazel_lib", version = "2.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.2")
 bazel_dep(name = "platforms", version = "0.0.8")
 bazel_dep(name = "helly25_bzl", version = "0.1.2")
 
+bazel_dep(name = "tar.bzl", version = "0.6.0")
+tar_toolchains = use_extension("@tar.bzl//tar:extensions.bzl", "toolchains")
+use_repo(
+    tar_toolchains,
+    "bsd_tar_toolchains_darwin_arm64",
```
fmeum: We would also want to support Intel Macs and potentially also Windows

toolchain/sysroot.bzl
```diff
@@ -0,0 +1,52 @@
...
+        cmd.extend(["--include", include])
+
+    for exclude in rctx.attr.exclude_patterns:
+        cmd.extend(["--exclude", exclude])
+
+    result = rctx.execute(cmd)
+    if result.return_code != 0:
+        fail(result.stdout + result.stderr)
+
+    rctx.delete(archive)
```
fmeum: Please also mark this repo rule as reproducible so that it benefits from those repo caches we manage to get into Bazel 9

fmeum (approved):

PR #586, opened 2025-10-26
Update llvm version used for tests.

helly25:
Allow attribute llvm_versions to specify version requirements.

Update some (main and test) dependencies.
Mark omp targets as manual (likely to be normal in a follow up).
Allow llvm_version to be taken from environment (or bazel command line). This does for newer bazel versions respect hermeticity.
Simplify version handling and log the resolution only in one place.
Update default llvm target version to "latest:>=17.0.0,<20" (e.g. 19) but allow environment overrides.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -81,7 +81,7 @@ build_test(
 dwp_file(
     name = "stdlib.dwp",
     src = ":stdlib_bin",
-    # NOTE: we should eventually we able to drop this; see #109.
+    # NOTE: we should eventually be able to drop this; see #109.
     override_compilation_mode = "dbg",
     target_compatible_with = [
         "@platforms//os:linux",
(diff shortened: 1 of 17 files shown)


fmeum (approved):

helly25:
@fmeum Please take another look.

tests/BUILD.bazel
```diff
@@ -101,6 +101,7 @@ cc_test(
     srcs = ["omp_test.c"],
     copts = ["-fopenmp"],
     linkopts = ["-fopenmp"],
+    tags = ["manual"],
```
fmeum: Should we run these explicitly on some platforms?

toolchain/internal/llvm_distributions.bzl
```diff
@@ -1159,24 +1168,57 @@ def _required_llvm_release_name(*, version_or_requirements, all_llvm_distributio
...
+        # We prefer 'repository_ctx.getenv' if it is available (~7.1+) and default
+        # to accessing the environment directly. The latter breaks "hermeticity".
+        if hasattr(rctx, "getenv"):
+            llvm_version = rctx.getenv(env_name, env_default)
+        elif env_name in rctx.os.environ:
+            llvm_version = rctx.os.environ[env_name]
+        else:
+            llvm_version = env_default
+
+        #print("\nINFO: Read environment variable '{env_var}' = '{llvm_version}'".format(
```
fmeum: Can we drop this?

tests/WORKSPACE
```diff
@@ -134,9 +137,10 @@ llvm_toolchain(
 )
 
 ## Toolchain example with system LLVM; tested in GitHub CI.
+# The llvm_version must match the version specified in .github/workflows/tests.yml: sys_paths_test
 llvm_toolchain(
     name = "llvm_toolchain_with_system_llvm",
-    llvm_versions = LLVM_VERSIONS,
+    llvm_versions = {"": "16.0.0"},
```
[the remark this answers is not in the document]
helly25: Done.

tests/BUILD.bazel
helly25: Done: There are not many though since they do not work with new llvm versions. But at least they need to be explicitly disabled.

PR #596, opened 2025-11-02
Switch from OpenSSL to BoringSSL for tests

helly25:
We should not test whether a specific variant of getting SSL works. Instead we can verify that one well supported one works.


fmeum (approved):
Glad to see this go away!

PR #610, opened 2025-11-08
Explain how version/requirement can be used from env vars.

helly25:


README.md
```diff
@@ -142,6 +142,34 @@ supporting `<`, `<=`, `>`, `>=`, `==`, `!=`. Examples:
...
+        "": "getenv(LLVM_VERSION,latest:>=17.0.0,<20)",
+        "darwin-x86_64": "15.0.7",  # Verify this works as opposed to using one version.
+    },
+)
+```
+
+In this example, MacOS x86 machines have their LLVM version hard-coded to
+`15.0.7`. For all other targets the LLVM version is read from the environment
+variable `LLVM_VERSION` which must be referenced on the Bazel command line as
+explained above. If the variabel is not present, then the LLVM version defaults
```
fmeum:
```suggestion
explained above. If the variable is not present, then the LLVM version defaults
```

fmeum (approved):

PR #616, opened 2025-11-13
Add load for cc_import

dzbarsky:


dzbarsky:
> @dzbarsky is it possible to get this backported to `v1.5.0` and cut a BCR release `1.5.0.bcr.1` or `1.5.1`? I have found that this is the only commit required to make `toolchains_llvm` functional in Bazel 9.0.0. Forcing an update to `v1.6.0` seem to break experimental docker worker sandboxing in a pretty profound way, so I'm blocked from updating at this point.

Can you just apply a patch or flip the Bazel autloads flag to include rules_cc for now? Are you an end user or an intermediate module?

PR #624, opened 2025-11-26
toolchain: Set `link_libs` rather than `link_flags`

phlax:
without this our TSAN/MSAN CI fails as it clobbers our setup to use the instrumented stdlib


phlax:
i can follow up later/tomorrow

PR #625, opened 2025-11-27
toolchain: Switch to `link_libs` for `libstdc++`

phlax:


dzbarsky (approved):
Thanks! Let's see if we break anyone :)

Not gonna block on this but it would be awesome to add a test with the sanitizer setup you were mentioning

phlax:
id like to offer to add tests - but realistically im unlikely to have spare cycles for this just now - also our san setup is kinda specific to our ci build pipeline so isolating a minimal reproducer is not straightforward

fwiw, the issue related to instrumented libc++ libs so not directly related to this pr - altho, i should say, we dont test clang+libstdc++ so this change is untested from our end

helly25:
If this gets merged I will see if more tests are necessary.

helly25 (approved):

phlax:
if you could hold off just a little - ill update to our latest patch - it renames the arg so keen to have update included

phlax:
oh wait .. i mixed this up with the other pr for x-compiling - will update that one today

PR #630, opened 2025-12-03
[bugfix] Reference os variable instead of os function

ha1dfo:
on toolchain/internal/common.bzl:121, we referenced the `os` function instead of the local `_os` variable, which made my local build throw stacktraces on toolchain_llvm v1.5.0 (was ok in v1.4.0).

I no longer have stacktraces with this fix locally present..


dzbarsky (approved):
thanks!

PR #655, opened 2026-01-18
Test again with Bazel 8 and include release candidates for testing

limdor:
The fix required for Bazel 8 was already merged

### DIFF
--- a/tests/transitions.bzl
+++ b/tests/transitions.bzl
@@ -19,6 +19,8 @@
 #
 # These three Bazel flags influence whether or not `.dwo` and `.dwp` are
 # created.
+load("@rules_cc//cc:defs.bzl", "CcInfo", "DebugPackageInfo")
+
 def _fission_transition_impl(settings, attr):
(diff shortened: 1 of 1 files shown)


limdor:
I will need some approval to trigger the tests. I tried to validate locally if with Bazel 8 works but I did not manage to make all tests run. If there is any information on how to run the tests locally let me know, I also commented about it in https://github.com/bazel-contrib/toolchains_llvm/issues/298#issuecomment-3765170564.

fmeum (commented):
Could you switch to `.*` instead of `.x` to include RCs and also add `9.*`?

limdor:
@fmeum it did not work with `*`, a too old version of bazelisk is currently being used. I just pushed another commit to upgrade it to `1.28.0` that is the latest. If you prefer another version let me know.

dzbarsky:
I think we also need DebugPackageInfo?

limdor:
> I think we also need DebugPackageInfo?

Yes, now it is added

Issue #694, opened 2026-03-09
Add mechanism for adding/fetching target libc++ libs for x-compile


phlax:
Currently the toolchain falls back to libstdc++ for any x-compile scenarios - iiuc on the basis that the host is more likely to have that available, and the libc++ in the toolchain won't work (wrong arch)

We (envoy) are currently hitting this for downstreams that want to x-compile to arm (cf https://github.com/envoyproxy/envoy/pull/43826), and im about to add a workaround that bundles a package containing the arm libs/bins and then exposes them via a repo rule

Ideally this is something that could be handled inside toolchains_llvm directly - perhaps using some similar mechanism to how sysroot is pulled in currently

there is a related macos issue here https://github.com/bazel-contrib/toolchains_llvm/issues/386

AustinSchuh:
https://github.com/bazel-contrib/toolchains_llvm/commit/a1ebc9aae6478820e358c53b2a2063edb1d430f0 works for this, also providing the libcompiler-rt for the target.

PR #696, opened 2026-03-09
[WIP] toolchain: Add cxx_lib attribute to inject libc++ in x-compile

phlax:
when cross-compiling you only need the exec toolchain and not the target one, except for the c++ libs themselves

to work around this currently toolchains_llvm falls back to using libstdc++ on the assumption that it is more likely to be present either on the host or in the sysroot

these foreign arch libs are only provided in the huge arch-specific toolchain or release tarballs, so we were keen not to have consumers downloading multiple toolchains for additional arches

to workaround this, we (envoy) now publish tarballs containining only `lib/` and `include/` folders with the libc++ libs fished out of the  llvm release tarball

this PR allows toolchains_llvm to make use of this - example config from current envoy (workspace):

```starlark

load("@envoy_toolshed//compile:libcxx_libs.bzl", "setup_libcxx_libs")
setup_libcxx_libs()

load("@toolchains_llvm//toolchain:rules.bzl", "llvm_toolchain")

llvm_toolchain(
    name = "llvm_toolchain",
    llvm_version = "18.1.8",
    # These libs are only included for cross-compile targets
    cxx_lib = {
        "linux-aarch64": "@libcxx_libs_aarch64",
        "linux-x86_64": "@libcxx_libs_x86_64",
    },
    cxx_standard = {"": "c++20"},
    sysroot = {} if USE_LOCAL_SYSROOT else {
        "linux-x86_64": "@sysroot_linux_amd64//:sysroot",
        "linux-aarch64": "@sysroot_linux_arm64//:sysroot",
    },
    toolchain_roots = {"": LLVM_PATH} if LLVM_PATH else {},
)
    
```

it can also be used with bzlmod like so

```starlark
llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm", dev_dependency=True)
llvm.cxx_lib(
    name = "llvm_toolchain",
    label = "@libcxx_libs_aarch64",
    targets = ["linux-aarch64"],
)
llvm.cxx_lib(
    name = "llvm_toolchain",
    label = "@libcxx_libs_x86_64",
    targets = ["linux-x86_64"],
)
llvm.toolchain(
    name = "llvm_toolchain",
    llvm_version = "18.1.8",
)
use_repo(llvm, "llvm_toolchain")
```
partial fix for #694

### DIFF
--- a/toolchain/extensions/llvm.bzl
+++ b/toolchain/extensions/llvm.bzl
@@ -73,17 +73,30 @@ def _llvm_impl_(module_ctx):
                 name,
             )
 
+            cxx_lib = {}
+            for cl in mod.tags.cxx_lib:
+                if cl.name != name:
+                    continue
+                for target in cl.targets:
+                    if cxx_lib.get(target):
+                        fail("duplicate target '%s' for cxx_lib with name '%s'" % (target, name))
+                    cxx_lib[target] = str(cl.label)
+            attrs["cxx_lib"] = cxx_lib
+
             llvm_toolchain(
                 **attrs
             )
 
-        # Check that every defined toolchain_root or sysroot has a corresponding toolchain.
+        # Check that every defined toolchain_root, sysroot, or cxx_lib has a corresponding toolchain.
         for root in mod.tags.toolchain_root:
             if root.name not in toolchain_names:
                 fail("toolchain_root '%s' does not have a corresponding toolchain" % root.name)
         for root in mod.tags.sysroot:
             if root.name not in toolchain_names:
                 fail("sysroot '%s' does not have a corresponding toolchain" % root.name)
+        for cl in mod.tags.cxx_lib:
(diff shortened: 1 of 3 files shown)


dzbarsky (commented):
What issues did you hit? In my experience, the c++ standard library `include/` doesn't vary between targets (aside from a few minor things like fortran and libomp headers) and when cross-compiling the cxx lib comes from the sysroot, which this toolchain already supports?

phlax:
fwiw - i think `cxx_lib` -> `cxx_cross_lib` to make it clear its only used for cross compile

more fixes from downstream testing being tested atm - will update once that is done

helly25:
@phlax should we merge as it's it still wait?

phlax:
apologies @helly25 would be good to land this but i need to make a couple of changes - our patch has iterated since - will follow up today/tomorrow

PR #703, opened 2026-03-24
Fix LLVM binary resolution for unrecognized Linux distros

zakcutner:
Unrecognized Linux distributions (e.g., Void Linux) failed with "No matching config could be found" because `_dist_to_os_names` returned an empty list, and both the aarch64 and x86 code paths had no useful fallback. The aarch64 path short-circuited on an empty list, and the x86 path fell back to the raw distro name (e.g., "void") which never matches any LLVM distribution.

Pass `["linux-gnu", "unknown-linux-gnu"]` as the default to `_dist_to_os_names` in both code paths so unrecognized distros fall back to generic linux-gnu binaries.


fmeum (approved):

marmeladema:
Thanks a lot for the fast merge! @zakcutner was actually kind to fix an issue impacting me :-)

When do you think a new release will be available?

PR #723, opened 2026-05-05
download_llvm: keep bundled-distro fallback reproducible

Wheest:
When a user sets `urls` on `llvm.toolchain(...)` for some platforms but not all, hosts on the unmatched platforms fall through to the bundled distribution table, and `download_llvm` then breaks reproducibility reporting in a way that disables `--repo_contents_cache` symlink-sharing for the LLVM dist.

Concretely, in `toolchain/internal/llvm_distributions.bzl::download_llvm`:

```python
if rctx.attr.urls:
    urls, sha256, strip_prefix, key = _urls(rctx)   # key="", urls=None on unmatched host
    if not sha256:
        update_sha256 = True                         # <- fires
if not urls:
    urls, sha256, strip_prefix = _distribution_urls(rctx)   # bundled table, has its own sha256
```

This means every output_base ends up with its own freshly-extracted ~7 GB copy of the LLVM dist rather than a symlink into the shared cache.

Fix is to clear `update_sha256` after the bundled fallback

Note/context: working in a monorepo, and as a heavy user of git worktrees, I've been finding disk bloat, as each worktree gets a copy of LLVM extracted.  I investigated this, and got a working patch internally.  To upstream, this is my first contrib to this repo, so worth noting in case I've misunderstood something

### DIFF
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -858,6 +858,7 @@ def download_llvm(rctx):
             update_sha256 = True
     if not urls:
         urls, sha256, strip_prefix = _distribution_urls(rctx)
+
         # The bundled distribution table provides its own sha256, so there
         # is nothing to record back into the rule's attrs. Leaving
         # update_sha256 set here would make `download_llvm` write
--- a/toolchain/internal/llvm_distributions.bzl
+++ b/toolchain/internal/llvm_distributions.bzl
@@ -858,6 +858,14 @@ def download_llvm(rctx):
             update_sha256 = True
     if not urls:
         urls, sha256, strip_prefix = _distribution_urls(rctx)
+        # The bundled distribution table provides its own sha256, so there
+        # is nothing to record back into the rule's attrs. Leaving
+        # update_sha256 set here would make `download_llvm` write
+        # `rctx.attr.sha256[""]` post-extract and put `llvm_repo_impl` on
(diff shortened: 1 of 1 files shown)


Wheest:
## Validation

Reproduced and validated downstream against a Bazel 8.6 monorepo, two fresh worktrees, same target on macOS arm64:

|                                    | Before               | After                                                       |
| ---------------------------------- | -------------------- | ----------------------------------------------------------- |
| `external/...llvm_toolchain_llvm`  | real per-output-base | symlink into `<repo_contents_cache>/<hash>/1`               |
| `bin/clang` inode across worktrees | different            | **same**                                                    |
| `external/` per output base        | 8.3 G                | **~750 M** (the LLVM dist is now shared)                    |

No change for users that supply both `urls` and `sha256` for their host platform — `_urls(rctx)` returns a real sha256 and `update_sha256` stays False as before.

helly25 (approved):

Wheest:
lint failure, should be sorted now

helly25:
Fantastic finding! Thanks a lot.

Wheest:
> Fantastic finding! Thanks a lot.

cheers, I was getting bored of clearing out 100s of GB of orphaned worktree output bases every week or so!

helly25:
@Wheest please sync again. Then I can approve running the workflow again and we should be good for merge.

Wheest:
> @Wheest please sync again. Then I can approve running the workflow again and we should be good for merge.

done @helly25 🫡

PR #726, opened 2026-05-09
ci: add stdlib matrix testing (libc++, stdc++, dynamic-stdc++)

helly25:
## Summary

Extend `external_test` in `.github/workflows/tests.yml` to cover all supported stdlib implementations by adding a `stdlib` matrix dimension and corresponding test toolchains.

### CI matrix (`.github/workflows/tests.yml`)

The `external_test` job now includes a `stdlib` dimension with four values:

| stdlib | behaviour |
|---|---|
| `default` | No override — auto-config picks `builtin-libc++` |
| `libc++` | System/sysroot-hosted libc++ |
| `stdc++` | Static libstdc++ |
| `dynamic-stdc++` | Dynamic libstdc++ |

**Exclusions to keep the matrix manageable:**

- **macOS + non-default stdlib** — libstdc++ is not available on macOS by default (would need homebrew install)
- **bzlmod=false + non-default stdlib** — only test the default `builtin-libc++` in legacy mode
- **bazel latest + bzlmod=false** — not needed

**Bazel versions:** restricted to `8.*` and `latest` (the old "rules_rust bzlmod support is experimental" comment from 2023 is stale).

### Test toolchains (`tests/MODULE.bazel` + `tests/WORKSPACE`)

Three new toolchain configurations added to both bzlmod and WORKSPACE modes:

- `llvm_toolchain_libcxx` — `stdlib = "libc++"`
- `llvm_toolchain_stdcpp` — `stdlib = "stdc++"`
- `llvm_toolchain_dynamic_stdcpp` — `stdlib = "dynamic-stdc++"`

### Test script (`tests/scripts/run_external_tests.sh`)

- **Platform detection** — uses `uname` to auto-detect the correct toolchain label suffix (`aarch64-darwin`, `x86_64-linux`, etc.) instead of hardcoding `x86_64-linux`
- **`TEST_STDLIB` env var** — selects which extra toolchain to pass via `--extra_toolchains`
- **`targets` array** — all test targets collected into a single array, passed as one `-- "${targets[@]}"` argument to `bazel test`
- **Conditional rust/go tests** — rules_rust and rules_go cgo prep+targets only run on bazel 9+ with bzlmod enabled (rules_rust bzlmod support is unreliable on older Bazel versions)

### WORKSPACE boringssl fix

Added missing `patch_args` and `patches` attributes to the boringssl `http_archive` in `tests/WORKSPACE` to match the bzlmod version in `tests/MODULE.bazel`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,9 +65,15 @@ jobs:
         bzlmod: [true, false]
         stdlib: [default, libc++, stdc++, dynamic-stdc++]
         exclude:
-          - bazel_version: latest # bazel latest does not need bzlmod=false
+          - os: macos-latest
+            stdlib: libc++ # Needs to be installed first
+          - os: macos-latest
+            stdlib: stdc++ # Needs to be installed first
+          - os: macos-latest
+            stdlib: dynamic-stdc++ # Needs to be installed first
+          - bazel_version: latest
             bzlmod: false
-          - bzlmod: false # bzlmod=false only tests the default (builtin-libc++) stdlib
+          - bzlmod: false # Only test the default stdlib: builtin-libc++
             stdlib: libc++
           - bzlmod: false
             stdlib: stdc++
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,17 +61,28 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest]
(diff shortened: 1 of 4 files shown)


helly25:
@fmeum we need to change the required tests since the two external tests now have an additional parameter. It is not clear which ones we want to require.

helly25:
@phlax take a look if the matrix for the external tests make sense to you please.

fmeum (approved):
Modifying required checks on GH is an unfortunate experience - I'll do my best

PR #731, opened 2026-05-14
Add --gcc-toolchain support

matt-sm:
In my company's setup we have an internal gcc toolchain.  We want to use the sysroot from this toolchain with clang builds via toolchains_llvm.

The only way I can reliably get this to work is to use `--gcc-toolchain` in conjunction with the `llvm.sysroot` extension tag, as there are files outside the `<gcc_toolchain_root>/<triple>/sysroot` folder that clang looks for.

I also need `--gcc-triple` as our gcc toolchain has a non-standard triple that seems to also cause clang problems.

In addition, we have libtool 3rd party builds that drop the `-gcc-toolchain` and `--gcc-triple` flags. The best way we've found to deal with these dropped flags is within the wrapper itself.

This PR adds bzlmod tags for gcc_toolchain and gcc_triple, threads them through repository generation, and restores them in cc_wrapper.sh when relink-like invocations drop them.

Included a regression test that exercises the new extension tags through the generated llvm toolchain repos.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -48,6 +48,17 @@ build_test(
     ],
 )
 
+build_test(
+    name = "gcc_toolchain_extension_test",
+    target_compatible_with = [
+        "@platforms//os:linux",
+    ],
+    targets = [
+        "@llvm_toolchain_gcc_context//:all-files-x86_64-linux",
+        "@llvm_toolchain_gcc_context//:cc-toolchain-x86_64-linux",
+    ],
+)
+
 # We want to test that `llvm-dwp` (used when assembling a `.dwp` file from
 # `.dwo` files) can be called.
 #
--- a/tests/MODULE.bazel
+++ b/tests/MODULE.bazel
@@ -105,10 +105,6 @@ llvm.gcc_toolchain(
     name = "llvm_toolchain_gcc_context",
     label = "//gcc_root:all_files",
 )
-llvm.gcc_triple(
-    name = "llvm_toolchain_gcc_context",
-    value = "x86_64-linux-gnu",
-)
 use_repo(llvm, "llvm_toolchain_gcc_context")
 
 # Toolchain examples for testing different stdlib implementations on Linux.
--- a/tests/MODULE.bazel
+++ b/tests/MODULE.bazel
@@ -32,8 +32,8 @@ bazel_dep(name = "abseil-cpp", version = "20250814.2", repo_name = "com_google_a
 bazel_dep(name = "boringssl", version = "0.20260508.0")
 single_version_override(
     module_name = "boringssl",
-    patches = ["//:boringssl-20260413.0.patch"],
     patch_strip = 1,
+    patches = ["//:boringssl-20260413.0.patch"],
 )
 
 # TODO: Remove when protobuf is released with a version of rules_python that supports 8.x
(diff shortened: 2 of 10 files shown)


helly25 (commented):
I don't fully get this. It appears you need to pass down a few flags which should be possible via .bazelrc and you also point to your linker which also can be done without patches. What is actually missing?

matt-sm:
> and you also point to your going

sorry what did you mean here

matt-sm:
If all I needed was a way to set those flags initially, then yes, I agree.

The gap I’m trying to solve is later wrapper-based relink invocations, especially via libtool. In that case the original Bazel action is configured correctly, but libtool reconstructs a new clang invocation and drops --gcc-toolchain / --gcc-triple. At that point .bazelrc or normal toolchain flag configuration is no longer enough, because the wrapper is being re-invoked with a reduced argv.

That is why I ended up looking at a llvm.gcc_toolchain / llvm.gcc_triple path together with wrapper fallback: not to introduce a new way to set the flags initially,  but to let the wrapper recover configured values when those relink invocations lose them.

If there is a cleaner way to preserve those flags through libtool relinks without wrapper fallback, I’m very open to that.

matt-sm:
I just noticed the rule has a test matrix back to llvm versions that don't support `--gcc-triple` so I would need to refactor and just keep `--gcc-toolchain` if this was viable

dzbarsky:
An alternative would be to rearrange your sysroot to behave how LLVM expects, right? Either statically or at build time via a repo rule ? Wouldn't that be simpler and not necessitate pushing complexity upstream?

matt-sm:
> rearrange your sysroot to behave how LLVM expects

I assume this is possible and is probably worth investigating.  It would need to be an arrangement that would not require either of these flags.  Every file LLVM requires would need to be discoverable from the sysroot.

Life would be simpler not maintaining a separate parallel gcc layout :)  It is unfortunate because `--gcc-toolchain` works with the current gcc toolchain as-is.

matt-sm:
I still think there is a valid use case here.

Even after removing the wrapper fallback and the flag-setting pieces from the patch, there is still a need to stage the GCC toolchain files and make them available through the generated toolchains_llvm filegroups. Without that, the GCC runtime pieces needed by Clang-backed toolchains are not available to Bazel actions in the expected way.

I’ve updated the patch to reduce it to that narrower behavior.

matt-sm:
Where this ended up, if there was an option like `build_data` that provisioned extra files required by the toolchain, that would probably be enough.

helly25:
> > and you also point to your going
> 
> sorry what did you mean here

lol, was done from the phone.

helly25:
> Where this ended up, if there was an option like `build_data` that provisioned extra files required by the toolchain, that would probably be enough.

@matt-sm Your update already looks cleaner and as @dzbarsky says you might just need to rearrange things. Maybe you can do that via a directory that builds up a link/directory structure that accomplishes that-similar to how Bazel actually works. I believe that there might indeed be a need to add additional files though. If the latter works for you, would you mind creating a separate PR for just that part and prevent unnecessary changes from reordering/reformatting.

helly25:
Another Q @matt-sm: have you looked at https://github.com/bazel-contrib/toolchains_llvm/pull/441 I assume it should at least make some of your goals easier.

AustinSchuh:
I've got a couple more changes once #441 lands which enable cross compiling against yocto sysroots, including with CUDA support.  We've got the same use case.

https://github.com/RealtimeRoboticsGroup/aos/tree/main/third_party/bazel-toolchain has a fully patched working version which includes all that.  I'll push up the rest (probably in a couple of smaller reviews) once #441 lands.

helly25:
#441 is now merged.

matt-sm:
I'm closing this in favour of https://github.com/bazel-contrib/toolchains_llvm/pull/749
               
The existing `extra_compiler_files` and `extra_linker_files` attrs do the job but are single-label with no per-target support. PR #749 adds `llvm.extra_compiler_files` and `llvm.extra_linker_files` bzlmod tag classes following the same pattern as `llvm.sysroot` — per-target, label-based, backwards compatible with the existing WORKSPACE attrs.

PR #743, opened 2026-05-30
Apply the path-prefix slash convention to sysroot_path

AustinSchuh:
The `*_path_prefix` arguments are required to end in "/" (asserted in cc_toolchain_config) and are concatenated with relative subpaths, e.g. `"{}lib/clang/...".format(prefix)`. `sysroot_path` did not follow this convention: it was never canonicalized, and usages were inconsistent -- some prepended a leading slash (`sysroot_path + "/usr/..."`) while the external_include_paths block did not (`sysroot_path + "usr/..."`). For an absolute sysroot without a trailing slash this produces a malformed path in the no-leading-slash sites; for a label/canonical sysroot (trailing slash) it produces a double slash in the leading-slash sites.

Make sysroot_path follow the same model:
- configure.bzl canonicalizes it to a directory path ending in "/".
- cc_toolchain_config.bzl asserts the trailing slash and concatenates relative subpaths everywhere (no leading slash).

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -318,6 +318,13 @@ def cc_toolchain_config(
     conly_flags = compiler_configuration["conly_flags"]
     sysroot_path = compiler_configuration["sysroot_path"]
 
+    # Follow the same convention as the `*_path_prefix` arguments: a non-empty
+    # sysroot path is a canonical directory path ending in '/', concatenated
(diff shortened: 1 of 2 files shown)


AustinSchuh:
The 3 failures are due to 503's from github.  I don't have permission to retest.

helly25 (approved):
For the moment the best we can do. Eventually we need a path join function.

PR #748, opened 2026-06-01
darwin: use the SDK's libc++; rework system include handling

AustinSchuh:
For a single-platform macOS build with the default `builtin-libc++` stdlib, Clang's driver auto-includes the toolchain's bundled libc++ headers (the LLVM version this toolchain was built from), while at link time libc++ is resolved from the macOS SDK sysroot. When those libc++ versions differ this is an ABI mismatch that can miscompile or crash at runtime (e.g. an undefined `std::__1::__hash_memory` at link).

Use the SDK's libc++ end to end on macOS: emit `-nostdinc++` and point `-cxx-isystem` at `<sysroot>/usr/include/c++/v1`, and exclude darwin from `use_toolchain_libcxx_paths` so the toolchain's libc++ header dirs are not added. Linking already pulls libc++/libc++abi from the sysroot.

To make that robust, the per-stdlib include handling is reworked into three accumulators that are turned into flags once, at the end:

- cpp_system_includes: C++ standard library header dirs, emitted as a single -nostdinc++ followed by -cxx-isystem entries.
- system_includes: compiler/C system header dirs, emitted as -idirafter.
- nostdinc: whether to also drop Clang's default C system include paths.

This centralizes the search-path ordering Clang requires (C++ headers via -cxx-isystem before C/system headers via -idirafter) instead of repeating it in each stdlib branch, and renames non_msan_compile_flags to toolchain_cpp_system_includes to reflect that it now holds header directories rather than ready-made flags.

Finally, fix linux stdc++-<ver> sysroot builds. They set -nostdinc, after which Clang no longer derives the sysroot's C system include dirs from --sysroot. Those dirs were computed (external_include_paths) but never added to the command line, so #include_next <stdlib.h> from libstdc++ wrappers like <cstdlib> could not reach the C library headers. They are now emitted as -idirafter after the compiler builtin headers, giving the search order: C++ headers -> compiler builtin -> C library.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -199,14 +199,16 @@ def cc_toolchain_config(
 
     major_llvm_version = int(llvm_version.split(".")[0])
 
-    external_include_paths = None
-
     resource_dir_version = llvm_version if major_llvm_version < 16 else major_llvm_version
 
     resource_dir = [
         "-resource-dir",
         "{}lib/clang/{}".format(target_toolchain_path_prefix, resource_dir_version),
     ]
+
+    # Clang's builtin headers (stddef.h, stdarg.h, intrinsics, ...). Searched
+    # after the C++ standard library headers (see system_includes below).
+    toolchain_builtin_include = target_toolchain_path_prefix + "lib/clang/{}/include".format(resource_dir_version)
 
     target_flags = [
         "--target=" + target_system_name,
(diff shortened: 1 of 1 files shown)


AustinSchuh:
One more after this which adds support for alternative sysroot layouts (yocto vs debian).  I'll add cross compilation tests to that one.

helly25 (approved):

PR #749, opened 2026-06-01
Add per-target extra_compiler_files and extra_linker_files bzlmod tags

matt-sm:
Adds llvm.extra_compiler_files and llvm.extra_linker_files tag classes to the bzlmod extension, following the same pattern as llvm.sysroot. Both tags accept a label and optional targets list, allowing per-target files to be made available in the toolchain's sandbox filegroups for compile and link actions respectively.

The existing single-label extra_compiler_files and extra_linker_files attrs on the WORKSPACE llvm_toolchain rule are unchanged.

### DIFF
--- a/tests/extra_files/BUILD.bazel
+++ b/tests/extra_files/BUILD.bazel
@@ -1,0 +1,5 @@
+filegroup(
+    name = "all_files",
+    srcs = ["ignorelist.txt"],
+    visibility = ["//visibility:public"],
+)
--- a/toolchain/extensions/llvm.bzl
+++ b/toolchain/extensions/llvm.bzl
@@ -21,12 +21,13 @@ def _root_dict(roots, cls, name, strip_target):
         for target in targets:
             if res.get(target):
                 fail("duplicate target '%s' found for %s with name '%s'" % (target, cls, name))
-            if bool(root.path) == (root.label):
+            path = getattr(root, "path", "")
+            if bool(path) == bool(root.label):
                 fail("target '%s' for %s with name '%s' must have either path or label, but not both" % (target, cls, name))
-            if root.path:
-                if not _is_absolute_path(root.path):
+            if path:
+                if not _is_absolute_path(path):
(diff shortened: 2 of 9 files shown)


toolchain/extensions/llvm.bzl
```diff
@@ -123,6 +133,22 @@ llvm = module_extension(
                 "path": attr.string(doc = "Absolute path to the sysroot."),
             },
         ),
+        "extra_compiler_files": tag_class(
+            attrs = {
+                "name": attr.string(doc = "Same name as the toolchain tag.", default = "llvm_toolchain"),
+                "targets": attr.string_list(doc = "Specific targets, if any; empty list means this applies to all."),
+                "label": attr.label(doc = "Label containing files to be made available in the sandbox for compile actions."),
```
helly25, line 141: Why do you need both label and path? Generally path dependencies do not work well throughout bazel and the label possibly already supports glob over 'dir/**'. If that works we should drop path. Same for the linker.

tests/extra_files/BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+filegroup(
+    name = "all_files",
+    srcs = ["dummy.txt"],
```
helly25: You want a test that actually verified your extra files make it all the way through to the compiler and linker. The latter is problematic but you can possibly test the former by bringing a source file with contents that can be tested.

toolchain/extensions/llvm.bzl
matt-sm: Yes I agree - I have dropped path

tests/extra_files/BUILD.bazel
matt-sm: I added a test that uses `-fsanitize-ignorelist` with an extra_compiler_file as the ignore file.  This will fail if the extra file is not passed through.

helly25 (approved):

PR #753, opened 2026-06-02
Support dynamic libstdc++ and Yocto sysroot layouts

AustinSchuh:
Add the last features needed to support yocto sysroots
- dynamic-stdc++ stdlib: a "dynamic-stdc++"[-<ver>] stdlib normalizes to the
  matching stdc++ configuration but links libstdc++.so instead of
  libstdc++.a (the static link is kept as the default). Removes the now-dead
  standalone "dynamic-stdc++" branch.

- multiarch override: a per-target `multiarch` attribute overrides the
  builtin multiarch tuple used to construct sysroot include/library paths,
  for sysroots that use non-standard tuples (e.g. Yocto aarch64-oe4t-linux).

- cxx_include_layout: a per-target `cxx_include_layout` attribute selects how
  libstdc++ headers and gcc runtime libs are laid out in the sysroot.
  `debian` (default) keeps the existing /usr/include/<multiarch>/c++/<ver>
  and /usr/lib/gcc/<multiarch>/<ver> layout; `yocto` uses
  /usr/include/c++/<ver>/<multiarch> and /usr/lib/<multiarch>/<ver>.

All three default to the prior behavior, so existing toolchains are
unaffected.


AustinSchuh:
Tests should all pass now.  I've got OSX -> linux x86, OSX -> linux arm64, linux x86 -> linux arm64, and linux x86 -> linux x86.  The bzlmod tests caught some missing plumbing which I added.

I couldn't find a good yocto rootfs to test against hosted in a decent spot.  (Mine are hosted on an internal site).  If anyone has ideas, I'd love to test against a public one.

AustinSchuh:
Retest (503)

toolchain/extensions/llvm.bzl
```diff
@@ -126,6 +131,14 @@ llvm = module_extension(
                 "path": attr.string(doc = "Absolute path to the toolchain root."),
             },
         ),
+        "target_toolchain_root": tag_class(
+            attrs = {
+                "name": attr.string(doc = "Same name as the toolchain tag.", default = "llvm_toolchain"),
+                "targets": attr.string_list(doc = "Specific targets, if any; empty list means this applies to all."),
+                "label": attr.label(doc = "Dummy label whose package path is the target toolchain root package."),
+                "path": attr.string(doc = "Absolute path to the target toolchain root."),
```
helly25: This is a new path attribute again.

.github/workflows/tests.yml
```diff
@@ -118,10 +118,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        os: [macos-latest, ubuntu-latest]
         bzlmod: [true, false]
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
```
helly25, line 123: Somehow the new tests do not schedule

toolchain/extensions/llvm.bzl
```diff
@@ -126,6 +131,14 @@ llvm = module_extension(
                 "path": attr.string(doc = "Absolute path to the toolchain root."),
             },
         ),
+        "target_toolchain_root": tag_class(
+            attrs = {
+                "name": attr.string(doc = "Same name as the toolchain tag.", default = "llvm_toolchain"),
```
helly25: All of this requires good documentation.

.github/workflows/tests.yml
AustinSchuh: I'm pretty sure they do.  Since the OS is also matrixed, they have a different "name", where the OS is part of the argument list now.  That is making the old "name" not run.  Happy to approach this differently if you'd like, but as is, I think there needs to be a corresponding change on the github configuration to point to the new name.

(Your other feedback is good, I'll get that to you hopefully soon)

toolchain/extensions/llvm.bzl
AustinSchuh: Thanks, done!  Feedback appreciated on the docs.

toolchain/extensions/llvm.bzl
AustinSchuh: I'm suspicious I don't understand what you are hinting at.  I added a bunch more docs.  Holler if that wasn't your intent.

.github/workflows/tests.yml
AustinSchuh: `Tests / xcompile_test (macos-latest, false) (pull_request)` is one of the new names.

AustinSchuh:
Failure looks like a download failure.  retest

.github/workflows/tests.yml
helly25: Yeah, I was blind. I set up https://github.com/bazel-contrib/toolchains_llvm/pull/757. Once that is in and configured we can get this PR merged automatically correct without worrying when we switch the requirements.

toolchain/extensions/llvm.bzl
helly25: Looks fine to me. Thanks!

toolchain/extensions/llvm.bzl
helly25: I don't like path arguments because bazel cannot handle them correctly. Hence the question: Can we do without?

AustinSchuh: I'll drop it.  It isn't required.  It is here to mirror the absolute paths in other part of `toolchains_llvm`.

helly25:
@AustinSchuh Once you merge in https://github.com/bazel-contrib/toolchains_llvm/pull/757 and we've fixed the config for the repo, you're renames should work just fine.

helly25:
@AustinSchuh: Merge to head. There we only check for the new `required_checks_done`. That verifies everything else passes. And there is a check to ensure everything is actually everything.

toolchain/extensions/llvm.bzl
AustinSchuh: I dropped it.

AustinSchuh:
retest (I swear posting this in the past has triggered failing tests to retry, might have been someone helping in the background)

helly25:
> retest (I swear posting this in the past has triggered failing tests to retry, might have been someone helping in the background)

Mostly me. Maybe something to implement...

PR #756, opened 2026-06-04
Redact date macros via -imacros header instead of -D on command line

matt-sm:
Reproducibility redaction of __DATE__/__TIME__/__TIMESTAMP__ was done with
-D__DATE__="redacted" etc. in unfiltered_compile_flags. The surrounding quotes
are required (the value is a string literal), but they do not survive being
passed through sub-build flag strings: tools that re-serialize the toolchain's
compile flags into their own build scripts (rules_foreign_cc make/cmake builds,
embedded compiler-flag strings) drop the quotes, leaving a bare `redacted`
token that the compiler treats as an identifier and fails on.

Write the three #define directives into a generated redacted_dates.h and
force-include it with `-imacros`, so no quoting has to survive the command
line.

Because layering_check requires every included header to be covered by a
module, the force-included header must appear in the generated system module
map. Add an `extra_textual_headers` input to system_module_map for individual
headers that must always be emitted as `textual header` entries -- unlike
cxx_builtin_include_files, whose entries are source directories emitted as
umbrella submodules (which cannot cover a single file). redacted_dates.h is
routed through this new input.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -41,6 +41,7 @@ def cc_toolchain_config(
         target_toolchain_path_prefix,
         tools_path_prefix,
         wrapper_bin_prefix,
+        redacted_dates_path,
         compiler_configuration,
         cxx_builtin_include_directories,
         extra_known_features,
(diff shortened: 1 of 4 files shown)


helly25 (approved):
Nice

helly25:
@matt-sm  Heads up: this was reverted in #825.

The `-imacros` header broke rules_rust build scripts (#771) and assembly compiles (#801). Both opt outs were rejected (#772, #774), so there was no way to keep it and unbreak those consumers.

The quoting problem you found is real and stays open. Toolchain flags that get re-serialized into a sub build command line lose the quotes. We think that belongs in the consumer, see #771.

Thanks for the original analysis!!! The write up here is what made the root cause clear.

PR #761, opened 2026-06-05
Build toolchain paths with helly25_bzl paths.join / ensure_trailing_slash

helly25:
## What

Replace the ad-hoc `+` / `.format()` path concatenations (which silently assumed every prefix ends in `/`) with `paths.join` / `paths.ensure_trailing_slash` from `@helly25_bzl//bzl/paths`, so path construction is explicit and slash-agnostic.

Also bumps the `helly25_bzl` **bazel_dep** to `0.4.3` in `MODULE.bazel`. #760 bumped only the WORKSPACE-mode `http_archive` (WORKSPACE / `toolchain/deps.bzl`); the bzlmod `bazel_dep` was left at `0.4.2`. 0.4.3 is required here because it adds `paths.ensure_trailing_slash`.

## Changes

- **`toolchain/cc_toolchain_config.bzl`** — tool paths, toolchain/sysroot include dirs, the `-resource-dir`/builtin-include paths, and the `-L`/`-B` flags now build via `paths.join`; `-B` and the yocto `-B`/`-Wl,-L` flags use `paths.ensure_trailing_slash` for their required directory slash. The `*_path_prefix` / `sysroot_path` "must end in `/`" assertions are replaced with `ensure_trailing_slash` normalization (guarantee instead of fail).
- **`toolchain/internal/configure.bzl`** — GCC C++ header autodetection, tool/lib symlink targets, and the dist-path prefix.
- **`toolchain/internal/common.bzl`** — `canonical_dir_path` → `paths.ensure_trailing_slash`; `pkg_path_from_label` → `paths.join` (now returns a canonical directory path, so callers don't need to wrap it).
- **`toolchain/internal/system_module_map.bzl`** — execroot-relative header paths via `paths.join`; relative include prefixes via `paths.ensure_trailing_slash`.

**Not changed:** Bazel *label* constructs (`Label(root + ":BUILD.bazel")`, `root + ":"`, `"@@…//"`) stay as string ops — `//` and `:` are label syntax that `paths.join`/`normalize` would corrupt.

## Testing

`tests/scripts/run_tests.sh` builds + tests pass from a clean (expunged) tree on macOS (fresh compiles, clang/resource-dir/include paths all resolve). The Linux libc++/stdc++/sysroot/yocto paths are covered by the CI matrix.


toolchain/internal/configure.bzl
```diff
@@ -235,9 +235,9 @@ def llvm_config_impl(rctx):
         tools_path_prefix = "bin/"
```
AustinSchuh, line 235: Drop the / here?

AustinSchuh (approved):

toolchain/internal/configure.bzl
helly25: We could possibly change that but we guarantee the trailing slashes for now and even require some of them, so imo consistency rules.

PR #765, opened 2026-06-07
Add MemorySanitizer support; drive all sanitizers via Bazel features

AustinSchuh:
This adds end-to-end MemorySanitizer (msan) support and moves AddressSanitizer/UndefinedBehaviorSanitizer/ThreadSanitizer off the `//toolchain/config` flags onto Bazel features, with an end-to-end msan CI job that exercises the instrumented libc++.

Why features instead of `//toolchain/config` flags --------------------------------------------------- Sanitizers must instrument the *target* program but must NOT instrument the build tools that run in the exec configuration (e.g. `flatc`, protoc plugins). A `bool_flag`/`config_setting` is a global build setting in bazel 8: it propagates into the exec configuration, so enabling it instruments the build tools too. Under msan that makes the tools abort at startup; under asan/ubsan it needlessly slows them and can break them.

Bazel resets `--features` to `--host_features` in the exec configuration, so a feature-enabled sanitizer is automatically dropped for build tools. That is the property we want, and it is why msan is wired through a `msan` cc_feature and why asan/ubsan/tsan now use rules_cc's stock `asan`/`ubsan`/`tsan` features (`--features=asan|ubsan|tsan|msan`). The old `//toolchain/config:{asan,ubsan}` flags are removed.

MemorySanitizer and the instrumented libc++
-------------------------------------------
msan reports false positives unless the C++ standard library is also instrumented, so msan swaps in an instrumented libc++ supplied via the new `libcxx_url`/`libcxx_sha256` attributes (extracted to `libcxx-msan/`). On Linux this is wired through mutually exclusive cc_features: the `msan` feature carries the instrumented-libc++ include/link flags, and an always-on `*_nomsan_stdlib` feature (gated `none_of [msan]`) carries the default stdlib flags, so enabling `--features=msan` swaps the standard library. msan is Linux-only.

The instrumented archives and headers must be available as toolchain inputs in BOTH filegroup layouts Bazel uses:

  * Bazel < 9 (no merkle_cache_v2): linker uses `lib_legacy`, compiler uses `include`.
  * Bazel >= 9 (merkle_cache_v2): linker uses `lib`, compiler uses `cxx_builtin_include`.

Pitfall (fixed here): if the overlay is added to only one variant, the build silently degrades instead of failing loudly:
  * A missing `libcxx-msan/lib/*.a` makes `-l:libc++.a` fall back to the *uninstrumented* base `libc++.a`. The link still succeeds, but msan then reports use-of-uninitialized-value deep inside libc++ (e.g. `std::ios_base::precision`) for any program that touches iostreams.
  * A missing `libcxx-msan/include/...` makes the instrumented headers absent from the compile sandbox (`'cstdio' file not found`). Both the `include`/`lib_legacy` and `cxx_builtin_include`/`lib` filegroups now carry the overlay, so msan works identically on Bazel 8 and 9.

Augmenting the stock sanitizer features
---------------------------------------
rules_cc's stock asan/ubsan/tsan features link via a plain `-fsanitize=...`, which does not pull Clang's C++ sanitizer runtime. C++ programs then fail to link (ubsan's vptr handlers: undefined `__ubsan_vptr_type_cache`, `__ubsan_handle_dynamic_type_cache_miss_abort`). We therefore augment each stock feature with `-fsanitize-link-c++-runtime`, and add ubsan's `-fsanitize=bounds` / `-fsanitize=nullability` (checks outside the default `undefined` group).

The augmentation is applied with a `select()` keyed on `config_setting`s that match `--features=asan|ubsan|tsan`. Keying on `--features` (rather than a `//toolchain/config` flag) keeps the exec-configuration reset, so the extra flags are dropped for build tools exactly like the stock features.

Rejected alternatives:
  * Defining our own cc_features named `asan`/`ubsan`/`tsan`: Bazel rejects this with "feature ... specified multiple times" because unix_cc_toolchain_config already emits them; overriding requires a `cc_external_feature` target.
  * Gating the extra flags on the stock features with `cc_external_feature` / `cc_feature_constraint`: `cc_external_feature` (and the provider file `cc_toolchain_info.bzl` it needs) are `visibility()`-restricted to `//cc/toolchains/...` in rules_cc (through 0.2.19), so they cannot be loaded from this repo, and the rule cannot be vendored either.
  * Adding the flags unconditionally: `-fsanitize=bounds`/`nullability` would instrument every build, and the flags would leak into the exec config.

Tests / CI
----------
  * `//:msan_libcxx_test` builds and runs an instrumented-libc++ program (the iostream/precision pattern above) with the `@llvm_toolchain_msan` toolchain and `--features=msan`. Unlike the analysis-only `msan_flags_test`, it links and runs, so it catches an uninstrumented libc++ being linked.
  * A dedicated `msan_test` CI job runs it across Bazel 7/8/9/latest so both filegroup layouts are exercised; it is wired into the `required_checks_done` gate.
  * `sanitizer_test` now enables sanitizers via `--features` and gains a `tsan_test`; asan/ubsan/tsan are verified to link and run end to end.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -163,6 +163,27 @@ jobs:
         env:
           USE_BZLMOD: ${{ matrix.bzlmod }}
         run: tests/scripts/run_tests.sh -t @llvm_toolchain_with_system_llvm//:cc-toolchain-x86_64-linux -v 16.0.0
+  msan_test:
+    # End-to-end MemorySanitizer test. Downloads a full clang + an
+    # instrumented-libc++ overlay and builds/runs an msan binary, exercising the
+    # instrumented-libc++ swap end to end. Linux/x86_64 and bzlmod only (the
+    # overlay toolchain is defined via bzlmod). The bazel-version matrix covers
+    # both filegroup layouts: <9 uses lib_legacy/include, 9+ uses
+    # lib/cxx_builtin_include (merkle_cache_v2).
+    strategy:
+      fail-fast: false
+      matrix:
+        bazel_version: [7.*, 8.*, 9.*, latest]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - run: tests/scripts/ubuntu_install_libtinfo.sh
+      - run: sudo rm -rf /usr/share/dotnet /usr/local/lib/android
+      - name: Test
+        env:
+          USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
+          USE_BZLMOD: true
+        run: tests/scripts/run_msan_tests.sh
   # Single aggregating gate for branch protection. It depends on every job that
   # must pass, so "required_checks_done" is the only status check that needs to
   # be marked "required" on the master branch -- adding or renaming a job above
@@ -180,6 +201,7 @@ jobs:
       - xcompile_test
       - abs_paths_test
       - sys_paths_test
+      - msan_test
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/README.md
+++ b/README.md
@@ -370,6 +370,70 @@ deps (also known as "depend on what you use") for `cc_*` rules. This feature
 can be enabled by enabling the `layering_check` feature on a per-target,
 per-package or global basis.
 
+### Sanitizers
+
+The toolchain can build with AddressSanitizer, UndefinedBehaviorSanitizer,
+ThreadSanitizer, or MemorySanitizer. Enable at most one at a time, via Bazel
+features:
+
+```sh
+bazel build //... --features=asan
+bazel build //... --features=ubsan
+bazel build //... --features=tsan
+bazel build //... --features=msan   # Linux only; see below
+```
+
+`asan`, `ubsan`, and `tsan` are rules_cc's stock sanitizer features and work
+with the regular standard library. Enabling sanitizers as features means Bazel
+resets them to `--host_features` in the exec configuration, so build tools stay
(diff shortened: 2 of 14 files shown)


helly25:
Very nice. Just silly findings to deal with/work around.

AustinSchuh:
It feels a bit weird for me to host the compilers for this test, but it should work fine.  If you've got a better spot, I'm all ears.

I feel decently comfortable with the strategies in use to get the flags into the right spot, but would love it if there is a cleaner way.  I tried to make sure there were enough comments and tests capturing the why for the next person.

dzbarsky:
> A bool_flag/config_setting is a global build setting: it propagates into the exec configuration, so enabling it instruments the build tools too.

This is no longer the case after Bazel 9. Flags have a `scope` argument and must explicitly opt into `universal` to propagate across the exec cfg transition.

BTW, we have [wired up all of this already a while ago in hermetic-llvm](https://github.com/search?q=repo%3Ahermeticbuild%2Fhermetic-llvm+msan&type=code) in case you want to compare notes

AustinSchuh:
> This is no longer the case after Bazel 9. Flags have a scope argument and must explicitly opt into universal to propagate across the exec cfg transition.

Ah, I'm stuck on bazel 8 for a while.  Too many workspace only dependencies left :(  But thanks, that's super helpful for the future.  I adjusted the description of the PR to be correct.

> BTW, we have wired up all of this already a while ago in hermetic-llvm in case you want to compare notes

Cool.  I took a quick look, nothing jumps out.  Fun to see independent people come up with similar answers.

Issue #184, opened 2023-02-03
missing dependency errors when building with `--copt -fsanitize=address` with `absolute_paths = True`


garymm:
Whenever I build anything with `--copt -fsanitize=address`, I get errors like:

```
ERROR: /home/garymm/.cache/bazel/_bazel_garymm/3e948c06be790e3e64e6aacacdab772a/external/libpng/BUILD.bazel:22:11: Compiling pngwio.c failed: undeclared inclusion(s) in rule '@libpng//:libpng':
this rule is missing dependency declarations for the following files included by 'pngwio.c':
  '/home/garymm/.cache/bazel/_bazel_garymm/3e948c06be790e3e64e6aacacdab772a/external/llvm_toolchain_llvm/lib/clang/15.0.6/share/asan_ignorelist.txt'
```

Would this be fixed if we add `"lib/clang/{}/share".format(llvm_version),` here?
https://github.com/grailbio/bazel-toolchain/blob/08d1aa8cee67a7a9c51c09225cead1dad42dadc3/toolchain/cc_toolchain_config.bzl#L271

siddharthab:
I don't mind the change you are suggesting. If you send a PR that works for you, I will accept it.

However, I can not reproduce this with the unit tests in this repo. Can you give me a reproducible example please? Can you also try your example with `--copt -fno-sanitize-ignorelist`?

siddharthab:
For example, in the unit tests here:

```
cd tests
bazel build --copt -fsanitize=address --linkopt -fsanitize=address @com_google_absl//...
```

garymm:
Ah it appears this happens only when `absolute_paths = True` is set in `llvm_toolchain()`.

siddharthab:
Thanks for looking into it.

PR #767, opened 2026-06-10
osx: load sanitizer runtime dylibs via @loader_path rpath (fixes #192)

### LINKED ISSUE #192: `--features=asan` doesn't work on macOS

helly25:
## Problem

Linking with `-fsanitize=address` (or ubsan/tsan) on macOS references the runtime as `@rpath/libclang_rt.<san>_osx_dynamic.dylib`, and Clang adds a *toolchain-relative* `LC_RPATH` (`external/.../lib/clang/<v>/lib/darwin`). That rpath is relative to the working directory, so it only resolves when cwd is the execroot — true at link time, but `bazel test` runs the binary from its runfiles, where the path doesn't exist:

```
dyld: Library not loaded: @rpath/libclang_rt.asan_osx_dynamic.dylib
```

This is why `--features=asan` doesn't work on macOS (#192). Workarounds people reach for don't hold up: `absolute_paths = True` makes the rpath absolute but breaks sandboxed builds (#184), and adding the toolchain libs as a `data` dep doesn't help under bzlmod (the rpath's `external/<canonical>` layout doesn't exist in bzlmod runfiles).

## Fix

After linking, if the output references a `@rpath/libclang_rt.*` dylib, add an `@loader_path`-relative `LC_RPATH` to the toolchain's compiler-rt `darwin` directory. `@loader_path` is relative to the binary rather than the working directory, so `@rpath` resolves wherever the binary runs — no `absolute_paths`, no copying the dylib into runfiles.

- Scoped to the hermetic (relative-prefix) toolchain; the host toolchain finds the runtime itself.
- Only fires when a sanitizer runtime is actually linked, so non-sanitized builds are unaffected.

## Testing

macOS 26 / Apple Silicon, LLVM 22.1.7: a downstream project's full AddressSanitizer suite (74 targets) passes with no `absolute_paths` and no data deps. (Note: LLVM ≤ 20 separately hangs in compiler-rt `FindDynamicShadowStart` on macOS 26, unrelated to this rpath fix.)

Fixes #192

### DIFF
--- a/toolchain/osx_cc_wrapper.sh.tpl
+++ b/toolchain/osx_cc_wrapper.sh.tpl
@@ -226,3 +226,30 @@ for rpath in ${RPATHS}; do
     fi
   done
 done
+
+# Sanitizer runtimes (e.g. libclang_rt.asan_osx_dynamic.dylib) are linked with an
+# `@rpath/<name>` install name, and Clang adds a *toolchain-relative* LC_RPATH
+# (`external/.../lib/clang/<v>/lib/darwin`) that only resolves when the process's
+# working directory is the execroot -- true at link time, but not when a test
+# runs from its runfiles, so the dylib is "not loaded" at runtime. Add an
+# `@loader_path`-relative LC_RPATH to the toolchain's compiler-rt darwin dir so
+# `@rpath` resolves regardless of the working directory -- no absolute paths
+# (which break the sandbox) and no copying the dylib into runfiles. Only for the
+# hermetic (relative-prefix) toolchain; the host toolchain finds it normally.
+if [[ ${toolchain_path_prefix} != /* ]] && [[ -n ${OUTPUT} && ${OUTPUT} != "1" && -f ${OUTPUT} ]]; then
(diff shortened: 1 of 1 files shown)


helly25:
Heads-up: #769 supersedes this PR — it contains this commit unchanged plus a `dynamic_runtime_lib`-based mechanism that ships the sanitizer runtime dylib in runfiles (covers remote execution and binaries moved out of `bazel-out`, which the rpath rewrite alone cannot). The two compose on disjoint files; full analysis and comparison in `docs/osx-sanitizer-runtime.md` over there. Happy to keep this one open as the minimal local-execution fix if reviewers prefer landing it separately.

helly25:
Replaced by 769

PR #772, opened 2026-06-13
configure: support TOOLCHAINS_LLVM_DISABLE_DATE_REDACTION env-var opt-out

### LINKED ISSUE #771: PR #756 -imacros redacted_dates.h breaks rules_rust cargo_build_script with --strategy=CargoBuildScriptRun=local
## Summary

PR #756  injects `-imacros external/toolchains_llvm++llvm+llvm_toolchain/redacted_dates.h` into `unfiltered_compile_flags`. The path is sandbox-relative, which is fine for normal `cc_library`/`cc_binary` actions but breaks consumers that re-serialize toolchain flags into out-of-sandbox compiler invocations.

The most common surface: `rules_rust` `cargo_build_script` actions invoking `cc-rs` when `--strategy=CargoBuildScriptRun=local` is set. `cc-rs` calls `cc_wrapper.sh` with the toolchain's flags but resolves CWD outside the sandbox where `external/...` paths don't exist.

## Reproducer

Any Rust target whose dep graph pulls a `*-sys` crate with a C build script (`ring`, `libtcmalloc-sys`, `snmalloc-sys`, etc.) plus:

```
# .bazelrc
build:linux --strategy=CargoBuildScriptRun=local
```

(`--strategy=CargoBuildScriptRun=local` is the recommended workaround for the `libtcmalloc-sys` sandbox-symlink issue, per `rules_rust` discussion.)

## Failure

```
cargo:warning=<built-in>:1:19: fatal error: 'external/toolchains_llvm++llvm+llvm_toolchain/redacted_dates.h' file not found
cargo:warning=    1 | #__include_macros "external/toolchains_llvm++llvm+llvm_toolchain/redacted_dates.h"
cargo:warning=      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cargo:warning=1 error generated.
```

Reproduces on darwin-aarch64 + linux-x86_64. `v1.8.0` (just released) carries PR #756, so every consumer that bumps past `v1.7.0` is affected.

## Cause

`toolchain/internal/configure.bzl:841`:

```python
redacted_dates_path = "external/{}/redacted_dates.h".format(rctx.name),
```

The sandbox-relative path is correct for in-sandbox actions but unusable for any consumer that doesn't have `cxx_builtin_include_files` in its runfiles. `rules_rust` `cargo_build_script_runner` doesn't propagate `cxx_builtin_include_files` to local-strategy invocations.

PR #756's docstring foresaw this: *"tools that re-serialize the toolchain's compile flags into their own build scripts (rules_foreign_cc make/cmake builds, embedded compiler-flag strings) drop the quotes"* — the new `-imacros` form doesn't have the quote-loss bug but instead has an analogous file-path-loss bug for the same class of consumers.

## Proposed fixes (pick one)

1. **Feature flag**: Add a \`disable_date_redaction\` toolchain feature (default off). Consumers can enable it to skip the `-imacros redacted_dates_path` block. Smallest API surface; opt-in; ships immediately.
2. **Bazel-resolved absolute path**: `redacted_dates_path = str(rctx.path("redacted_dates.h"))`. Bakes machine-specific path into config — breaks remote execution cache.
3. **Restore -D__DATE__="redacted"**: revert PR #756. Trades cargo_build_script bug for rules_foreign_cc quote-loss bug.
4. **Bazel runfile path propagation**: cooperate with rules_rust to propagate `cxx_builtin_include_files` to local-strategy cargo_build_script. Out of scope here.

Happy to send a PR implementing option 1 (`disable_date_redaction` feature) if maintainers want.

## Environment

- bazel 7.5.0
- toolchains_llvm \`332d6856\` (current HEAD, v1.8.0 + 4 commits)
- rules_rust 0.70.0
- LLVM 21.1.8 / 20.1.8 (darwin-aarch64)
- macOS 26.4 + Ubuntu 24.04 Linux

jayakasadev:
Fixes #771.

## Problem

PR #756 injects `-imacros external/.../redacted_dates.h` into `unfiltered_compile_flags`. The path is sandbox-relative. That breaks any consumer whose actions re-serialize toolchain flags into out-of-sandbox compiler invocations:

- `rules_rust` `cargo_build_script` with `--strategy=CargoBuildScriptRun=local` (the recommended workaround for the `libtcmalloc-sys` sandbox-symlink issue, per rules_rust discussion). Every `-sys` crate that ships C source fails with `fatal error: '.../redacted_dates.h' file not found`.
- `rules_foreign_cc` make/cmake builds that embed compiler-flag strings — the original failure mode PR #756 was trying to fix; the new form has the analogous file-path-loss bug.

Reproducer + full context in #771.

## Fix

Smallest viable: env-var opt-out at toolchain configure time. `TOOLCHAINS_LLVM_DISABLE_DATE_REDACTION=1` (or `true`) → `redacted_dates_path` is emitted empty → the `-imacros` injection block is skipped in `unfiltered_compile_flags`.

**Default behavior unchanged.** Existing consumers see no impact.

## Why env var, not feature flag or attribute

- Module extension attribute would need plumbing through `extensions/llvm.bzl` + `internal/repo.bzl` + `internal/configure.bzl` — wider API surface.
- A Bazel `cc_feature` would mean a runtime feature toggle, but `unfiltered_compile_flags` is applied at toolchain config time in a different layer from the feature pipeline. Bigger refactor.
- Env var at `repository_ctx` time is exactly the right granularity — the decision happens once when the toolchain is configured.

## Alternatives considered

- **Absolute path via `rctx.path("redacted_dates.h")`** — bakes a machine-specific path into the cc_toolchain config; breaks remote execution cache.
- **Revert PR #756, restore `-D__DATE__="redacted"`** — trades one bug for another.
- **Propagate `cxx_builtin_include_files` to local-strategy cargo_build_script** — requires cooperation from rules_rust; out of scope here.

Happy to expand to a full feature-flag / module-attribute if maintainers prefer that shape.

## Files

- `toolchain/internal/configure.bzl` — read env var at repo rule time; emit empty path when opted-out.
- `toolchain/cc_toolchain_config.bzl` — guard the `-imacros`/`-Wno-builtin-macro-redefined` injection on `redacted_dates_path` being non-empty.

## Test plan

- [x] Default build unchanged (env var unset → original injection present).
- [x] With env var set, `bazel build` of a Rust target whose dep graph includes `ring` build script succeeds (verified locally on darwin-aarch64 + linux-x86_64 against `toolchains_llvm@332d6856`).
- [ ] CI matrix coverage — happy to add if maintainers want a regression test.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -224,13 +224,16 @@ def cc_toolchain_config(
     unfiltered_compile_flags = [
         # Do not resolve our symlinked resource prefixes to real paths.
         "-no-canonical-prefixes",
-        # Reproducibility: redact date macros via an -imacros header rather than
-        # -D__DATE__="redacted" on the command line, whose quotes are lost when
-        # passed through sub-build flag strings.
-        "-Wno-builtin-macro-redefined",
-        "-imacros",
-        redacted_dates_path,
     ]
+    if redacted_dates_path:
+        unfiltered_compile_flags += [
+            # Reproducibility: redact date macros via an -imacros header rather than
+            # -D__DATE__="redacted" on the command line, whose quotes are lost when
+            # passed through sub-build flag strings.
+            "-Wno-builtin-macro-redefined",
+            "-imacros",
+            redacted_dates_path,
+        ]
 
     major_llvm_version = int(llvm_version.split(".")[0])
 
--- a/toolchain/internal/configure.bzl
+++ b/toolchain/internal/configure.bzl
@@ -838,7 +838,14 @@ cc_toolchain(
         target_toolchain_path_prefix = target_toolchain_path_prefix,
         tools_path_prefix = toolchain_info.tools_path_prefix,
         wrapper_bin_prefix = toolchain_info.wrapper_bin_prefix,
-        redacted_dates_path = "external/{}/redacted_dates.h".format(rctx.name),
+        # Env-var opt-out for consumers whose actions re-serialize toolchain
+        # flags into out-of-sandbox compiler invocations (rules_rust
+        # cargo_build_script with --strategy=CargoBuildScriptRun=local,
+        # rules_foreign_cc make/cmake). The sandbox-relative redacted_dates.h
+        # path cannot resolve outside the sandbox.
+        redacted_dates_path = (
+            "" if rctx.os.environ.get("TOOLCHAINS_LLVM_DISABLE_DATE_REDACTION", "") in ("1", "true", "True") else "external/{}/redacted_dates.h".format(rctx.name)
+        ),
         sysroot_label_str = sysroot_label_str,
         sysroot_path = sysroot_path,
         stdlib = stdlib,


helly25:
I think users would want to be able to control this via their .bazelrc file, so they can enable and disable this per config. So we would need more support than the env var.

fmeum:
An env var opt-out also isn't great since it's all or nothing. Could this be gated behind a feature instead?

dzbarsky:
I left my thoughts on the linked issue: https://github.com/bazel-contrib/toolchains_llvm/issues/771#issuecomment-4698929438

PR #774, opened 2026-06-14
Make date-macro redaction a toggleable feature (fixes #771)

### LINKED ISSUE #771: PR #756 -imacros redacted_dates.h breaks rules_rust cargo_build_script with --strategy=CargoBuildScriptRun=local

helly25:
## Problem

PR #756 (shipped in **v1.8.0**) redacts `__DATE__`/`__TIME__`/`__TIMESTAMP__` for reproducibility by force-including a generated `redacted_dates.h` via `-imacros`, placed in the always-on `unfiltered_compile_flags`. The header path is **sandbox-relative**, so any consumer that re-serializes the toolchain's compile flags and runs the compiler **outside** Bazel's sandbox fails with `fatal error: '.../redacted_dates.h' file not found`:

- `rules_rust` `cargo_build_script` with `--strategy=CargoBuildScriptRun=local` (every `*-sys` crate that ships C source)
- some `rules_foreign_cc` make/cmake builds

This is a regression for anyone upgrading past v1.7.0. See #771.

## Approach

As noted on #771, the root cause is in the consumers (they leave the sandbox), so this repo's job is to offer a clean, **configurable** opt-out — not the all-or-nothing env-var of #772. Per the discussion (control via `.bazelrc`, gate behind a feature), this exposes the redaction as a standard Bazel **`date_redaction` feature, enabled by default** so behavior is unchanged out of the box.

The `-Wno-builtin-macro-redefined` / `-imacros redacted_dates.h` flags move out of the baked `unfiltered_compile_flags` into a `cc_args` + `cc_feature` wired on via `extra_enabled_features` — modeled on the existing msan/nomsan `cc_feature` pattern in the same file.

Consumers that need it disable it like any other feature:

```sh
bazel build //... --features=-date_redaction       # global
```
```python
cc_binary(name = "x", features = ["-date_redaction"])  # per-target
```
```
common --features=-date_redaction                   # .bazelrc
```

(Build-tool/exec-config actions reset `--features` to `--host_features`, so `--host_features=-date_redaction` covers those — documented in the README.)

## Tests

Two analysis-only tests (modeled on `msan_flags_test`):

- `date_redaction_default_test` — asserts `redacted_dates.h` is on the `CppCompile` command line by default.
- `date_redaction_disabled_test` — asserts it is **absent** under `--features=-date_redaction`.

Verified locally (darwin-arm64, LLVM 19.1.7): `aquery` shows the flags present by default and gone with the feature disabled; both new tests and the existing `stdlib_test` pass; buildifier clean.

## Files

- `toolchain/cc_toolchain_config.bzl` — `date_redaction` cc_feature (default-on).
- `tests/transitions.bzl`, `tests/BUILD.bazel` — regression tests.
- `README.md` — "Date-macro redaction" section.

Fixes #771. Supersedes the env-var approach in #772.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ cc_binary(
 )
 ```
 
-```
+```text
 # Or per-config in .bazelrc.
 common --features=-date_redaction
 ```
--- a/README.md
+++ b/README.md
@@ -362,6 +362,46 @@ then they can be referenced as:
 - `@llvm_toolchain//:clang-format`
 - `@llvm_toolchain//:llvm-cov`
 
+### Date-macro redaction
+
+For reproducible builds the toolchain redacts the `__DATE__`, `__TIME__`, and
+`__TIMESTAMP__` preprocessor macros, replacing them with a constant. This is
+done by force-including a generated header with `-imacros`, which is wired
+through the `date_redaction` feature that is **enabled by default** — no
+configuration is needed for the common case.
+
+A small number of consumers re-serialize the toolchain's compile flags and run
+the compiler outside Bazel's sandbox — for example `rules_rust`
+`cargo_build_script` with `--strategy=CargoBuildScriptRun=local`, or some
+`rules_foreign_cc` make/cmake builds. There the sandbox-relative path of the
+force-included header cannot be resolved and the compile fails. For those cases
+the redaction can be turned off like any other Bazel feature, on a per-target,
+per-package, or global/`.bazelrc` basis:
+
+```sh
+# Disable globally (note the leading "-", which means "disable the feature").
+bazel build //... --features=-date_redaction
+```
+
+```python
+# Or disable for a single target.
+cc_binary(
+    name = "uses_local_cargo_build_script",
+    features = ["-date_redaction"],
+    # ...
+)
+```
+
+```
+# Or per-config in .bazelrc.
+common --features=-date_redaction
+```
+
+Disabling it only drops the redaction flags; nothing else about the toolchain
+changes. Note that build-tool (exec-configuration) actions reset `--features`
+to `--host_features`, so to also disable redaction for them use
+`--host_features=-date_redaction`.
+
 ### Strict header deps (Linux only)
 
 The toolchain supports Bazel's `layering_check` feature, which relies on
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -15,7 +15,7 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
-load(":transitions.bzl", "dwp_file", "msan_flags_test", "sanitizer_test", "transition_library_to_platform")
+load(":transitions.bzl", "date_redaction_default_test", "date_redaction_disabled_test", "dwp_file", "msan_flags_test", "sanitizer_test", "transition_library_to_platform")
 
 cc_library(
     name = "stdlib",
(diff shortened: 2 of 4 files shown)


helly25:
We decided against this feature.

Issue #532, opened 2025-08-11
Consider adding nostdlib++ whenever a stdlib is explicitly indicated


jwbee:
This toolchain adds enough flags to cause programs to be linked to libc++ and libc++abi, in which case the build will succeed with the `-nostdlib++` flag. The flag will also prevent the eventuality that a program is linked to both libc++ and libstdc++. Consider adding it to the toolchain flags where appropriate.

A related safety improvement would be to add nostdinc++ whenever the libc++ header paths are being given as flags, to prevent searching for (and erroneously finding) the GNU c++ headers.

PR #781, opened 2026-06-20
fix(toolchain): drop redundant -stdlib=libc++ that triggers -Wunused

guille2718:
PR #748 reworked libc++ header handling so the bundled headers are located via explicit -cxx-isystem entries plus a single -nostdinc++ (emitted by _cxx_isystem whenever cpp_system_includes is non-empty). With -nostdinc++ on the compile line, -stdlib=libc++ no longer affects anything at compile time (it only steers the libc++ header search path, which is now overridden), so clang correctly reports it unused, printing once per object file:

    clang: warning: argument unused during compilation: '-stdlib=libc++'
    [-Wunused-command-line-argument]

It is also unnecessary at link: libc++ is linked explicitly via -l:libc++.a / -l:libc++abi.a (Linux) or -lc++ / -lc++abi (Darwin), and cxx_flags feed compile actions only, never the link action.

Guard -stdlib=libc++ so it is only passed when -nostdinc++ will NOT be emitted (i.e. cpp_system_includes is empty), mirroring the existing stdc++-<version> branch which already omits -stdlib for the same reason. The only config that still needs it is stdlib=libc++ on Darwin, where cpp_system_includes stays empty and clang's auto-detection requires it.

### DIFF
--- a/toolchain/cc_toolchain_config.bzl
+++ b/toolchain/cc_toolchain_config.bzl
@@ -451,9 +451,6 @@ def cc_toolchain_config(
         stdlib = "stdc++-" + stdlib[len("dynamic-stdc++-"):]
 
     if stdlib == "builtin-libc++":
-        cxx_flags.extend([
-            "-stdlib=libc++",
-        ])
         system_includes.append(toolchain_builtin_include)
         if is_darwin_exec_and_target:
             # On macOS, use the SDK's libc++ entirely (headers + linking).
@@ -511,10 +508,6 @@ def cc_toolchain_config(
             ]
 
     elif stdlib == "libc++":
-        cxx_flags.extend([
-            "-stdlib=libc++",
-        ])
-
         stdlib_link_libs.extend([
             "-l:libc++.a",
             "-l:libc++abi.a",
@@ -602,6 +595,16 @@ def cc_toolchain_config(
     use_toolchain_libcxx_paths = stdlib in ["builtin-libc++", "libc++"] and target_os != "darwin"
     if use_toolchain_libcxx_paths:
         cpp_system_includes = toolchain_cpp_system_includes
+
+    # -stdlib=libc++ only affects the C++ header search path at compile time,
+    # which -nostdinc++ overrides (emitted by _cxx_isystem whenever
+    # cpp_system_includes is non-empty). libc++ is linked explicitly via
+    # -l:libc++.a / -lc++, and cxx_flags never reach the link action. So only
+    # pass it when we are NOT emitting -nostdinc++; otherwise it produces a
+    # spurious -Wunused-command-line-argument on every compile (cf. the
+    # stdc++-<version> branch, which omits -stdlib for the same reason).
+    if stdlib in ["builtin-libc++", "libc++"] and not cpp_system_includes:
+        cxx_flags.append("-stdlib=libc++")
 
     if major_llvm_version >= 14:
         # With C++20, Clang defaults to using C++ rather than Clang modules,


dzbarsky:
Is this just a warning fix? The choice of cxx lib is quite complex unfortunately - even for a single OS you might need to link different ones depending if you are building a dynamic library to be linked into a prebuilt (such as python/node extension) or a full cc binary, etc. in my experience using this toolchains in a few companies, this cxx lib flag ends up being patches to do whatever is needed. I suspect the cases I've seen are not isolated, so that means your PR will cause conflicts to many downstream users and force them to update their patches, etc. Is it worth it?

We just had another recent example of a contribution that tweaked some edge case behavior but broke other edge cases, so I'd prefer to minimize churn if it's not necessary to improve correctess/etc

guille2718:
Yes, this is just a warning fix. But the current code produces a warning on every single one of the outputs in my repo, and disabling the warning is not something I think is a good idea.

Happy to think harder about edge cases, but I'd be surprised to be the only one for which the new version of llvm_toolchains started producing a lot of warnings...

mikael-s-persson:
I'm hitting that issue too. And I suspect everyone using libc++ is too.

I like the fix. The only thing I feel a bit uncomfortable with is the placement of the conditional append of the "-stdlib=libc++" flag. It's gated on `stdlib in ["builtin-libc++", "libc++"] and not cpp_system_includes`, but the `cpp_system_includes` list is not committed to being part of the final flags until much later, when it is transformed using `default_cxx_isystem_flags = _cxx_isystem(cpp_system_includes)`. I think the code is correct, but brittle because a future code-change could break this assumption (even if it's well documented) that if `cpp_system_includes` is not empty around line 600 it still won't be empty by line 730 or so, or even later when `default_cxx_isystem_flags` is actually used.

I think it would be better to effectively do it at the same time and place where `_cxx_isystem` is used and bundle that option within `default_cxx_isystem_flags`. I think that makes that particular flag bundle a more self-contained "replace the builtin includes" set of options, rather than having that "fix" in the more general `cxx_flags` bundle.

FYI: As the maintainer of [bazel_cc_meta](https://github.com/mikael-s-persson/bazel_cc_meta) that has to do a lot of hacky reverse-engineering of these arcane inclusion path rules, and as a user of many weird cross-compilation / sysroot toolchains (including some "hand-rolled" armv7 toolchains), I'm gonna try to test this out soon'ish (I'm mostly concerned about the underlying PR #748 that caused this issue).

helly25:
Alternative approach in #791: rather than guarding `-stdlib=libc++` behind `not cpp_system_includes`, it removes the flag from the `builtin-libc++`/`libc++` branches entirely — it's redundant in every config (moot under `-nostdinc++`, and darwin already defaults to libc++) and never reaches the link action. #791 also adds a `//:stdlib_flags_test` regression guard, and includes evidence that `-nostdlib++` (#532) isn't needed here (no libstdc++ in the binary's dynamic `NEEDED`, in either opt or fastbuild). Cross-referencing for coordination — happy to go with whichever approach the maintainers prefer.

guille2718:
I actually  prefer @helly25's solution over mine, I think it is cleaner. I don't mind which solution we take exactly though, just that the warnings become signal again :)

helly25:
> Yes, this is just a warning fix. But the current code produces a warning on every single one of the outputs in my repo, and disabling the warning is not something I think is a good idea.
> 
> Happy to think harder about edge cases, but I'd be surprised to be the only one for which the new version of llvm_toolchains started producing a lot of warnings...

@guille2718 can you confirm on https://github.com/bazel-contrib/toolchains_llvm/pull/791 whether it works for you or not?

helly25:
@guille2718 https://github.com/bazel-contrib/toolchains_llvm/pull/791 has been merged.

guille2718:
@helly25 Yes! This works. Thanks for the fix!

PR #791, opened 2026-07-01
fix(toolchain): drop redundant -stdlib=libc++ that clang reports unused

helly25:
## Problem

Since #748 reworked libc++ header handling, the toolchain locates the libc++ headers explicitly via `-nostdinc++` + `-cxx-isystem`. With `-nostdinc++` on the compile line, `-stdlib=libc++` no longer affects header selection, and since it never reaches the link action (`cxx_flags` feed compile actions only), it is entirely unused — clang reports it once per object file:

```
clang: warning: argument unused during compilation: '-stdlib=libc++' [-Wunused-command-line-argument]
```

## Fix

Remove `-stdlib=libc++` from the `builtin-libc++` and `libc++` branches entirely, rather than conditionally guarding it (cf. #781). It is redundant in every configuration:

- Wherever `-nostdinc++` is emitted (builtin-libc++ on Linux and macOS, libc++ on Linux) it has no compile-time effect.
- On darwin `libc++` (the one config without `-nostdinc++`) clang already defaults to libc++.
- libc++ is linked explicitly via `-l:libc++.a` / `-l:libc++abi.a` (Linux) or `-lc++` / `-lc++abi` (Darwin); `-stdlib` never reaches the link action.

This is an alternative to #781: rather than guarding a redundant flag, it removes the flag at the source.

### Why not also add `-nostdlib++` (#532)?

#532 additionally suggests emitting `-nostdlib++` so a program can never link both libc++ and libstdc++. I checked empirically (LLVM 19.1.7, Linux): the prebuilt binaries do **not** double-link libstdc++ — no `libstdc++` in the binary's dynamic `NEEDED` in either `-c opt` or fastbuild. That concern does not manifest in this toolchain, so adding `-nostdlib++` would be churn without a correctness benefit. Left out of scope.

## Test

Adds `//:stdlib_flags_test`, an analysis-only test asserting the C++ compile command line carries `-nostdinc++` and **not** `-stdlib=libc++`. It introspects the `CppCompile` action's `argv` (like the existing `msan_flags_test`) and runs as part of `//:all` — i.e. the `test` CI job on Ubuntu and macOS, across every Bazel version and bzlmod setting.

## Validation

Verified on macOS (arm64) and Linux (arm64 container):

- `stdlib_flags_test` passes on both.
- A fresh C++ compile no longer emits the warning.
- libc++ programs build and run.
- The untouched `stdc++` path still builds and runs on Linux. (`stdc++` is unsupported on macOS — no system libstdc++, already excluded from the CI matrix — and is unchanged by this PR.)

Refs #781, #748, #532.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -15,7 +15,7 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
-load(":transitions.bzl", "dwp_file", "msan_flags_test", "sanitizer_test", "transition_library_to_platform")
+load(":transitions.bzl", "dwp_file", "msan_flags_test", "sanitizer_test", "stdlib_flags_test", "transition_library_to_platform")
 
 cc_library(
     name = "stdlib",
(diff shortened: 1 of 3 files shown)


mikael-s-persson:
I just tested this PR with a few setups I have, and it all seems to work as expected. I also checked the compiler_commands files extracted from the actions. All looks good. Those setups most pull in llvm-17 toolchains.

PR #795, opened 2026-07-04
feat(toolchain): add libxml2_stub attribute for prebuilt Linux ld.lld

oharboe:
## Problem

Prebuilt LLVM release binaries for Linux dynamically link `ld.lld` against `libxml2.so.2`, which lld only uses for Windows COFF manifest merging. libxml2 2.14 changed its soname to `libxml2.so.16`, so on hosts that ship only the new library (Ubuntu 25.10+, Arch, Fedora 41+) `ld.lld` cannot start:

```
ld.lld: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory
```

and on hosts with an unversioned compatibility symlink, every link action prints:

```
ld.lld: /lib/x86_64-linux-gnu/libxml2.so.2: no version information available (required by ld.lld)
```

Upstream context: llvm/llvm-project#138225. LLVM release binaries statically link libxml2 as of llvm/llvm-project commits 70cf763a42d5 and 4d7c1c6b08c0; the first release carrying the change is 23.1.0 (scheduled 2026-08-25). All earlier LLVM releases keep the dynamic dependency permanently.

## Change

New optional repo/toolchain attribute `libxml2_stub` (default `False`). When set, a minimal ELF64 shared object defining the sixteen symbols `ld.lld` imports — tagged with the `LIBXML2_2.4.30` / `LIBXML2_2.6.0` version definitions it requests — is written to `lib/libxml2.so.2` in the downloaded distribution. The binaries find it via their `RUNPATH` (`$ORIGIN/../lib`) ahead of any system search path, so they start and run silently with no host libxml2 anywhere. The symbols are never called on ELF targets; function bodies are trap instructions and `xmlFree` is a NULL pointer.

The bytes are assembled in pure Starlark and written with `rctx.file(..., legacy_utf8 = False)`, following the existing `libclang_rt` overlay pattern. Nothing is executed to produce the file, so fetching works identically from any host OS (including fetching a Linux toolchain from macOS or under remote execution). x86_64 and aarch64 are supported; the attribute is a no-op for non-Linux distributions.

## Verification

- Generated bytes are byte-identical (sha256) to a reference implementation, for both architectures.
- On Ubuntu 26.04 with no usable system libxml2: `ld.lld --version` fails without the stub, runs silently with it; `readelf -V` shows the version definitions; `nm -D` shows the versioned symbols; a `cc_binary` compiles, links and runs through the toolchain with a clean log.
- `tests:libxml2_stub_test` (Linux) asserts the stub exists in the distribution and that `ld.lld` starts with no libxml2 diagnostics; the main test toolchain enables the attribute so existing tests exercise link actions through it.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -242,6 +242,23 @@ sh_test(
     ],
 )
 
+# The libxml2 stub (`libxml2_stub = True` in tests/MODULE.bazel) must let
+# ld.lld start without any host libxml2.so.2. Linux-only: the stub is not
+# written into non-Linux distributions.
+sh_test(
+    name = "libxml2_stub_test",
+    srcs = ["libxml2_stub_test.sh"],
+    args = [
+        "$(rootpath @llvm_toolchain_llvm//:bin/ld.lld)",
+        "$(rootpath @llvm_toolchain_llvm//:lib/libxml2.so.2)",
+    ],
+    data = [
+        "@llvm_toolchain_llvm//:bin/ld.lld",
+        "@llvm_toolchain_llvm//:lib/libxml2.so.2",
+    ],
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
 # As a workaround for https://github.com/bazelbuild/rules_foreign_cc/issues/1018.
 toolchain(
     name = "ninja_mac_arm64_toolchain",
--- a/tests/MODULE.bazel
+++ b/tests/MODULE.bazel
@@ -73,6 +73,7 @@ LLVM_VERSIONS = {
 llvm.toolchain(
     name = "llvm_toolchain",
     cxx_standard = {"": "c++17"},
+    libxml2_stub = True,
     llvm_versions = LLVM_VERSIONS,
 )
 llvm.extra_target_compatible_with(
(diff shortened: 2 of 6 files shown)


dzbarsky:
While it's cool that you implement the stub in starlark, I think this is a little crazy :) Here's an alternative solution to this kind of problem: https://github.com/hermeticbuild/rules_rs/pull/139

In general, you might be interested in giving https://github.com/hermeticbuild/hermetic-llvm a spin, as it provides fully statically linked compilers that sidestep this completely

oharboe:
@dzbarsky Thank you for the feedback! I'm ditching this approach and pivoting to your suggestion. Let's see what that does...

oharboe:
Thanks — https://github.com/hermeticbuild/hermetic-llvm's statically linked binaries make the stub unnecessary, so closing this. For the record, for anyone landing here with the same problem (prebuilt LLVM release ld.lld requiring libxml2.so.2 on hosts that only ship libxml2.so.16): this PR's head commit remains consumable as an archive_override on toolchains_llvm, and the underlying issue is fixed upstream in LLVM release binaries starting with 23.1.0 (scheduled 2026-08-25, llvm/llvm-project commits 70cf763a42d5 / 4d7c1c6b08c0).

PR #799, opened 2026-07-08
Set `canonical_id` on `sysroot` download

eloisekp:
The `sysroot` rule doesn't pass `get_default_canonical_id()` to `rctx.download()`. Without it, the repo cache is keyed on the checksum alone, so changing a sysroot URL can serve stale cached bytes instead of refetching. This mirrors `http_archive` & `http_file` rules so a URL change forces a real download.

See https://github.com/bazelbuild/bazel/issues/22652#issuecomment-2151543618

### DIFF
--- a/toolchain/sysroot.bzl
+++ b/toolchain/sysroot.bzl
@@ -1,5 +1,5 @@
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
-load("@bazel_tools//tools/build_defs/repo:cache.bzl", "get_default_canonical_id")
+load("@bazel_tools//tools/build_defs/repo:cache.bzl", "DEFAULT_CANONICAL_ID_ENV", "get_default_canonical_id")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "get_auth")
 
 def _sysroot_impl(rctx):
@@ -71,6 +71,7 @@ def _sysroot_impl(rctx):
 
 sysroot = repository_rule(
     implementation = _sysroot_impl,
+    environ = [DEFAULT_CANONICAL_ID_ENV],
     attrs = {
         "url": attr.string(),
         "urls": attr.string_list(),
--- a/toolchain/sysroot.bzl
+++ b/toolchain/sysroot.bzl
@@ -1,4 +1,5 @@
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+load("@bazel_tools//tools/build_defs/repo:cache.bzl", "get_default_canonical_id")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "get_auth")
 
 def _sysroot_impl(rctx):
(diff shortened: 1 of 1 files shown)


toolchain/sysroot.bzl
rrbutani, line 75: would you mind listing `BAZEL_HTTP_RULES_URLS_AS_DEFAULT_CANONICAL_ID` as an env var dep here? (see [here](https://github.com/bazelbuild/bazel/blob/93867c22f19fbe945b139d71a7d72c1aa9d43928/tools/build_defs/repo/cache.bzl#L25) and [here](https://github.com/bazelbuild/bazel/blob/93867c22f19fbe945b139d71a7d72c1aa9d43928/tools/build_defs/repo/http.bzl#L633))

```suggestion
    environ = [DEFAULT_CANONICAL_ID_ENV],
    attrs = {
```

(plus a corresponding load)

dzbarsky: i thought that isn't needed if you use `rctx.getenv`?

dzbarsky: oh, lol. https://github.com/bazelbuild/bazel/blob/93867c22f19fbe945b139d71a7d72c1aa9d43928/tools/build_defs/repo/cache.bzl#L62

we should probably fix bazel...

dzbarsky: @fmeum is there a reason you didn't use `getenv` there? did it just not exist yet?

toolchain/sysroot.bzl
```diff
@@ -1,4 +1,5 @@
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+load("@bazel_tools//tools/build_defs/repo:cache.bzl", "get_default_canonical_id")
```
rrbutani: ~~FWIW I think `cache.bzl` ~~is `private`~~ is not present in Bazel <8~~

~~(imo this is fine; maybe worth calling it out in the release notes or something though)~~

EDIT: please disregard; I missed that both the change adding `cache.bzl` and the change making it public were backported to 7.x; Bazel 7.4.0+ should continue to work fine

toolchain/sysroot.bzl
fmeum: It existed, but didn't track the dependency. Yeah, let's fix Bazel.

rrbutani: https://github.com/bazelbuild/bazel/pull/30189; thanks @fmeum 

---

does it make sense to still apply the proposed change here, so we do the right thing w/older Bazel versions? (or is the intent to backport https://github.com/bazelbuild/bazel/pull/30189 to Bazel 7/8/9?)

fmeum: I think it does, the backport will take a while and may be too noisy (it changes all lockfiles)

eloisekp: Thanks all, done: https://github.com/bazel-contrib/toolchains_llvm/pull/799/changes/0075790fb7792c10d2a515c3f4de01413b9ea5dc

dzbarsky (approved):

toolchain/sysroot.bzl
dzbarsky: thank you!

dzbarsky (approved):

PR #801, opened 2026-07-09
Fix compiling assembly on OSX

AustinSchuh:
PR #756 introduced date macro redaction by force-including a generated redacted_dates.h header via the `-imacros` flag in unfiltered compile flags. Since unfiltered compile flags are passed to all driver invocations (including assembly compilations), clang tries to load this header when compiling assembler (`.s` / `.S`) files.

However, redacted_dates.h was only declared as an input file in the C/C++ compiler-components filegroup, and was missing from the assembler-files filegroup. This caused assembly compiles to fail with a fatal error:
  "external/toolchains_llvm++llvm+llvm_toolchain/redacted_dates.h file not found"
as the header was absent from the assembler action's sandbox.

Fix this by adding redacted_dates.h to the assembler-files filegroup definitions (both absolute-path and relative-path configurations) in toolchain/internal/configure.bzl.

Also add a regression test compiling an empty assembly file to ensure the assembler can successfully resolve redacted_dates.h.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -313,3 +313,14 @@ cc_test(
     tags = ["manual"],
     deps = [":test_cxx_standard_lib_transitioned"],
 )
+
+cc_library(
+    name = "assembly_test_lib",
+    srcs = ["assembly_test.s"],
+)
+
+build_test(
+    name = "assembly_test",
+    targets = [":assembly_test_lib"],
+)
+
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -323,4 +323,3 @@ build_test(
     name = "assembly_test",
     targets = [":assembly_test_lib"],
 )
-
--- a/tests/assembly_test.s
+++ b/tests/assembly_test.s
@@ -1,0 +1,1 @@
+# An empty assembly file to test that assembler action works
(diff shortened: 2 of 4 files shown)


fmeum:
@AustinSchuh We may have to roll back the original fix since it broke some other ways in which foreign build systems were run in a Bazel build.

helly25:
> @AustinSchuh We may have to roll back the original fix since it broke some other ways in which foreign build systems were run in a Bazel build.

Right. It's the reason there is no new release

AustinSchuh:
Sounds good.

I think the compile test should get merged anyways since I confirmed it caught the problem.  But happy to wait until the dust settles.

AustinSchuh:
Did this go anywhere?

helly25:
> Did this go anywhere?

@AustinSchuh Sorry for the late reply. Yes, it went somewhere. **#756 was reverted in #825**. That is what this was waiting for.

Your fix stays. The revert only drops `redacted_dates.h` from the `assembler-files-*` filegroups because the header is not generated anymore. The assembly test stays too. It now guards a path that has no bug anymore, but that is where you want a test anyway. Thanks for confirming it caught the regression.

We decided the `-imacros` approach broke more than it fixed. It broke rules_rust build scripts (#771) and assembly here. Both opt outs were rejected (#772, #774). So there was no way to keep it and unbreak the affected consumers. Redaction goes back to `-D__DATE__="redacted"`.

AustinSchuh:
OSX and Linux now work with no patches.  Thanks!

PR #824, opened 2026-08-16
Fix illegal ambiguous match when combining sanitizers (fixes #777)

### LINKED ISSUE #777: [regression from 1.7.0] Illegal ambiguous match on configurable attribute "link_flags"
This happens when renovate tries updating `toolchains_llvm` from `1.7.0` to `1.8.0`.

See the [failed build](https://github.com/DoDoENT/bazel-playground/actions/runs/27680764025/job/81867366048?pr=48).

The problem is that both `--features=asan` and `--features=ubsan` are given. Disabling either or both resolves the issue. However, this is not desirable in my scenario.

For convenience, here is the error from failed build:

```
(09:56:19) ERROR: /home/runner/.bazel/external/toolchains_llvm++llvm+llvm_linux_intel_haswell/BUILD.bazel:63:20: Illegal ambiguous match on configurable attribute "link_flags" in @@toolchains_llvm++llvm+llvm_linux_intel_haswell//:local-x86_64-linux:
@@toolchains_llvm+//toolchain/config:use_asan
@@toolchains_llvm+//toolchain/config:use_ubsan
Multiple matches are not allowed unless one is unambiguously more specialized or they resolve to the same value. See https://bazel.build/reference/be/functions#select.
```

helly25:
Fixes #777.

## Problem

Enabling two sanitizers at once fails during analysis:

```
$ bazel build //:stdlib_test --features=asan --features=ubsan
ERROR: .../llvm_toolchain/BUILD.bazel:194:13: Illegal ambiguous match on
configurable attribute "dynamic_runtime_lib" in
@@toolchains_llvm++llvm+llvm_toolchain//:cc-clang-aarch64-darwin:
@@toolchains_llvm+//toolchain/config:use_asan
@@toolchains_llvm+//toolchain/config:use_ubsan
```

`//toolchain/config:use_asan`, `:use_ubsan` and `:use_tsan` match on
`values = {"features": ...}`, so they are **not** mutually exclusive. With two
sanitizers on, two branches of the same `select()` match, and Bazel rejects an
ambiguous match unless one condition is more specialized or all matches resolve
to the same value.

The regression came in with the v1.8.0 sanitizer work (#765, #769) and is live
in the v1.8.0 that BCR serves. It shows up on `link_flags` on Linux (as in the
issue report) and on `dynamic_runtime_lib` on macOS. CI missed it because every
sanitizer test enables exactly one sanitizer.

## Fix

Make every sanitizer `select()` key on exactly one condition.

- New `//toolchain/config:use_any_sanitizer` (`config_setting_group`,
  `match_any`) for what all sanitizers share: the `-fsanitize-link-c++-runtime`
  link flag and the darwin runfiles feature. ubsan's extra
  `-fsanitize=bounds` / `-fsanitize=nullability` keep their own `:use_ubsan`
  select and are concatenated.
- `dynamic_runtime_lib` takes a single label, so it can't be built by
  concatenation. It now points unconditionally at a
  `sanitizer-runtime-libs-{suffix}` filegroup whose `srcs` are the
  concatenation of one single-condition `select()` per sanitizer, so a combined
  build gets exactly the runtime dylibs it enabled:

  ```python
  filegroup(
      name = "sanitizer-runtime-libs-aarch64-darwin",
      srcs = select({"...:use_asan": ["...:libclang_rt-asan-darwin"], "//conditions:default": []}) +
             select({"...:use_ubsan": ["...:libclang_rt-ubsan-darwin"], "//conditions:default": []}) +
             select({"...:use_tsan": ["...:libclang_rt-tsan-darwin"], "//conditions:default": []}),
  )
  ```

The darwin runfiles-feature select previously listed all three conditions with
identical values, which was legal only under the "resolve to the same value"
rule. Keying it on the group removes that subtlety.

## Tests

- `sanitizer_test`'s `sanitizer` string attr becomes a `sanitizers` list; new
  `//:asan_ubsan_test` builds and runs the combination end to end on Linux, and
  is listed in `tests/scripts/run_tests.sh` next to the existing sanitizer
  tests. asan+tsan is not covered because clang rejects that pair outright.
- New analysis-only `//:sanitizer_combo_flags_test` enables asan and ubsan
  together. Analysis-only means no sanitizer runtime is needed, so it also runs
  on macOS -- the only platform where the `dynamic_runtime_lib` ambiguity
  appears.

Verified locally on macOS (aarch64): `bazel build //:stdlib_test
--features=asan --features=ubsan` fails on master with the error above and
succeeds with this change; `//:sanitizer_combo_flags_test` fails with that same
error when the toolchain changes are reverted and passes with them.

This is one of the two regressions blocking the next release.

## What about the other sanitizers?

Worth stating explicitly, since "which sanitizers does this cover" is the
obvious question:

| Sanitizer | Driven by | Affected by #777? | Covered by the group? |
| --- | --- | --- | --- |
| asan | `:use_asan` config_setting | yes | yes |
| ubsan | `:use_ubsan` config_setting | yes | yes |
| tsan | `:use_tsan` config_setting | yes | yes |
| msan | `msan`/`nomsan` cc_features | no | no, by design |
| lsan | rules_cc stock feature, not augmented here | no | no -- gap, see below |

**msan** cannot hit this bug: it is carried entirely by the mutually-exclusive
`msan`/`nomsan` cc_features in `cc_toolchain_config.bzl`, so no `select()` keys
on it and there is no `:use_msan` setting to be ambiguous with anything. It
also does not want what the group guards -- it links the instrumented libc++
itself (`-l:libc++.a`, `-rtlib=compiler-rt`) rather than through
`-fsanitize-link-c++-runtime`, and it is Linux-only, so the macOS runtime
dylibs never apply. `--features=msan --features=ubsan` works after this change
for the same reason it is unaffected before it.

**lsan** is the one worth flagging, and it turned out to be a real bug rather
than a naming nit. rules_cc ships four stock sanitizer features -- asan,
**lsan**, tsan, ubsan -- and this toolchain only has config_settings for three,
so `--features=lsan` gets no augmentation. On macOS that is not cosmetic: an
lsan binary links against `@rpath/libclang_rt.lsan_osx_dynamic.dylib`
(confirmed with `otool -L`), and with no `:use_lsan` setting nothing ships that
dylib into runfiles -- the same failure #769 fixed for the other three.

Wiring lsan up changes behaviour for lsan builds and wants its own test, so I
have deliberately left it out of this PR rather than widen a release-blocking
regression fix, and recorded it where the group is defined. Happy to open the
follow-up.

The group is therefore named `use_common_sanitizer` -- for the treatment it
applies, not its current membership, since lsan should join it and the LLVM
distributions already carry an `rtsan` runtime besides. Both the msan exclusion
and the lsan gap are documented at the definition.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -15,7 +15,7 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
-load(":transitions.bzl", "dwp_file", "msan_flags_test", "sanitizer_test", "stdlib_flags_test", "transition_library_to_platform")
+load(":transitions.bzl", "dwp_file", "msan_flags_test", "sanitizer_combo_flags_test", "sanitizer_test", "stdlib_flags_test", "transition_library_to_platform")
 
 cc_library(
     name = "stdlib",
@@ -92,7 +92,7 @@ cc_test(
 sanitizer_test(
     name = "asan_test",
     src = ":stdlib_bin",
-    sanitizer = "asan",
+    sanitizers = ["asan"],
     tags = ["manual"],
     target_compatible_with = ["@platforms//os:linux"],
 )
@@ -100,7 +100,7 @@ sanitizer_test(
 sanitizer_test(
     name = "ubsan_test",
     src = ":stdlib_bin",
-    sanitizer = "ubsan",
+    sanitizers = ["ubsan"],
     tags = ["manual"],
     target_compatible_with = ["@platforms//os:linux"],
 )
@@ -108,7 +108,21 @@ sanitizer_test(
 sanitizer_test(
     name = "tsan_test",
     src = ":stdlib_bin",
-    sanitizer = "tsan",
+    sanitizers = ["tsan"],
+    tags = ["manual"],
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
+# asan and ubsan together: the combination that used to fail analysis with
+# "Illegal ambiguous match" (#777). asan+tsan is not tested because clang
+# rejects that pair outright.
+sanitizer_test(
+    name = "asan_ubsan_test",
+    src = ":stdlib_bin",
+    sanitizers = [
+        "asan",
+        "ubsan",
+    ],
     tags = ["manual"],
     target_compatible_with = ["@platforms//os:linux"],
 )
@@ -119,6 +133,15 @@ sanitizer_test(
 msan_flags_test(
     name = "msan_flags_test",
     target_compatible_with = ["@platforms//os:linux"],
+    target_under_test = ":stdlib_bin",
+)
+
+# Analysis-only check that two sanitizers can be enabled at once. Not restricted
+# to Linux and not tagged manual: it never links or runs, so it needs no
+# sanitizer runtime, and macOS is where the combination used to fail on the
+# cc_toolchain's dynamic_runtime_lib (#777).
+sanitizer_combo_flags_test(
+    name = "sanitizer_combo_flags_test",
     target_under_test = ":stdlib_bin",
 )
 
--- a/tests/scripts/run_tests.sh
+++ b/tests/scripts/run_tests.sh
@@ -76,7 +76,7 @@ if [[ -z "${toolchain_name}" ]]; then
     # and listed here only for the default toolchain: they require a sanitizer
     # runtime new enough for the host glibc, which the pinned LLVM 13.0.0
     # toolchain used by the container tests is not.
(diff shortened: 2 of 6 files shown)


fmeum (approved):

helly25:
Filed #826 for the lsan gap this turned up: on macOS an `--features=lsan` binary references `@rpath/libclang_rt.lsan_osx_dynamic.dylib` (confirmed with `otool -L`) and nothing ships it into runfiles, which is the same failure #769 fixed for the other three. Kept out of this PR on purpose — it changes behaviour for lsan builds and wants its own test, and this one is a release blocker.

helly25:
Correction on the lsan note above: I tested it before proposing a fix and the dyld failure I claimed does not reproduce — `--features=lsan` behaves identically with and without the wiring on a local macOS build, because the sanitizer dylibs are already reachable in the execroot. Details and the corrected scope are in #826. The inconsistency (no `:use_lsan`, so lsan gets none of the augmentation) is real, but it is not a release blocker and #824 is unaffected either way.

PR #839, opened 2026-08-28
Pass `DEVELOPER_DIR` env var to macOS SDK detection for better cache invalidation

mmirgttam:
On macOS, if a sysroot is not provided, we fall back to calling `xcrun`, which supports using the `DEVELOPER_DIR` environment variable to locate the SDK. This PR adds that env var to the `xcrun` invocation to force Bazel to refresh the toolchain repository when the value changes. 

I tested this with Bazel 9.1 by switching `DEVELOPER_DIR` while reusing the same output base. Before this change, Bazel reused the SDK path from the first run (!). After this change, changing `DEVELOPER_DIR` made it rerun repository configuration. And setting it to an invalid path forced an `xcrun` error, as expected.

There's some prior art here:
* https://github.com/bazel-contrib/rules_python/pull/3203
* https://github.com/bazelbuild/bazel/blob/9ed07e33051696b29302015e823ea8f697e7b160/tools/osx/xcode_configure.bzl#L304-L315
* https://github.com/bazelbuild/apple_support/blob/872de7149cfbcaaf854c7867e98b6d4798ba0d27/README.md#toolchain-configuration
* https://github.com/bazelbuild/rules_apple/blob/3a5bbed353c226f318bb7e5c99ac2330dd80bc85/doc/common_info.md#xcode-version-selection-and-invalidation


mmirgttam:
The one failing check seems to be an intermittent network issue:
```
History:
 - Timeout exceeded when accessing 'http://cdn.opensuse.org/tumbleweed/repo/oss/repodata/42b2c7e89318851bf7cb9ca652ae1c3081ff60d0613753253464ca75ce9711369f4e212ad511383641ca799ed51ca2b7417f688ff9bf286bb57184484a278b73-appdata.xml.gz'.
   Timeout reached Curl error (28)
```

dzbarsky:
Where are you seeing `xcrun`? This module provides a hermetic toolchain so it should not invoke xcrun. Those other examples you link are other rulesets trying to support non-hermetic toolchains

mmirgttam:
> Where are you seeing `xcrun`? This module provides a hermetic toolchain so it should not invoke xcrun. Those other examples you link are other rulesets trying to support non-hermetic toolchains

Here:
https://github.com/bazel-contrib/toolchains_llvm/blob/1207730cdb02e34a789b385d97279ec0605c3836/toolchain/internal/sysroot.bzl#L26

I believe this is the path taken when no sysroot is defined.

dzbarsky:
> > Where are you seeing `xcrun`? This module provides a hermetic toolchain so it should not invoke xcrun. Those other examples you link are other rulesets trying to support non-hermetic toolchains
> 
> Here:
> 
> https://github.com/bazel-contrib/toolchains_llvm/blob/1207730cdb02e34a789b385d97279ec0605c3836/toolchain/internal/sysroot.bzl#L26
> 
> I believe this is the path taken when no sysroot is defined.

I see, I didn't realize people were using that codepath, I think it's unfortunate. But in any case I think the proper fix is to `rctx.getenv` in that branch to get the watch to install only when that branch is taken. Otherwise the env var changing will invalidate the repo rule even with a sysroot provided, degrading the performance for everyone else

mmirgttam:
> > > Where are you seeing `xcrun`? This module provides a hermetic toolchain so it should not invoke xcrun. Those other examples you link are other rulesets trying to support non-hermetic toolchains
> > 
> > 
> > Here:
> > https://github.com/bazel-contrib/toolchains_llvm/blob/1207730cdb02e34a789b385d97279ec0605c3836/toolchain/internal/sysroot.bzl#L26
> > 
> > I believe this is the path taken when no sysroot is defined.
> 
> I see, I didn't realize people were using that codepath, I think it's unfortunate. But in any case I think the proper fix is to `rctx.getenv` in that branch to get the watch to install only when that branch is taken. Otherwise the env var changing will invalidate the repo rule even with a sysroot provided, degrading the performance for everyone else

Oh good point! I'll move it there.

mmirgttam:
I've updated it to only set the var in the no-sysroot fallback branch that calls `xcrun`, and updated the PR title and description to match.

helly25:
The idea seems correct.

Preferably we wait with meeting until we get the last missing 23.1.0 distribution added, so I can do the next release before this change (without release branch shenanigans).

dzbarsky (approved):
maybe leave a 1-line comment explaining that we're doing this for the watch side-effect?

mmirgttam:
> maybe leave a 1-line comment explaining that we're doing this for the watch side-effect?

Sure thing, done.

helly25:
@mmirgttam I sent this to auto-merge. Next release should follow shortly once we have the missing 23.1.0 distribution for MacOs.

helly25 (approved):

helly25:
@mmirgttam there is an issue with the PR as we are requiring signed commits. Do you think you can sign the commits?

mmirgttam:
> @mmirgttam there is an issue with the PR as we are requiring signed commits. Do you think you can sign the commits?

Yes, done! Should we add that requirement to the contributing docs?

PR #845, opened 2026-09-02
Add dependency-safe LLVM distribution tag

helly25:
## Motivation

Rulesets may need headers or static libraries from an official LLVM distribution without owning the consuming workspace's C++ toolchain selection. Today the bzlmod extension rejects every tag from a non-root module, so such rulesets either duplicate the distribution downloader or require every consumer to recreate their repository.

This is the narrower dependency-safe alternative to allowing dependency modules to configure complete toolchains in #568. It avoids the unresolved question of which module owns toolchain registration and aliases.

## Changes

- add `llvm.distribution(...)`, backed by the existing `llvm` repository rule
- permit that tag in dependency modules while keeping every toolchain/configuration tag root-only
- normalize `llvm_version` exactly as the combined `llvm_toolchain` macro does
- document the API, generated repository layout, and globally unique naming requirement

The tag only downloads/materializes the distribution. It does not create or register a C++ toolchain.

## Validation

- a two-module Bazel 9.2.0 integration fixture successfully built a root alias to `@distribution_dep//:clang`, where the dependency module alone declared `llvm.distribution(name = "distribution_dep_llvm", llvm_version = "22.1.8")`
- focused distribution tests pass:
  - `//toolchain/internal:llvm_distributions_output_test`
  - `//toolchain/internal:llvm_distributions_select_test`
  - `//toolchain/internal:llvm_prerelease_test`
  - `//toolchain/internal:llvm_requirements_test`
- `buildifier` and `git diff --check` pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,25 @@ The toolchain is tested to work with `rules_go`, `rules_rust`, and
 
 ### Accessing tools
 
+Dependency modules that need files from a matching LLVM distribution, but do
+not own the root module's C++ toolchain selection, can materialize only the
+distribution:
+
+```python
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.distribution(
+    name = "my_rule_llvm",
+    llvm_version = "22.1.8",
+)
+use_repo(llvm, "my_rule_llvm")
+```
+
+The generated repository uses the same `BUILD.llvm_repo.tpl` layout as the
+`<toolchain-name>_llvm` repository created by `llvm.toolchain`. It neither
+creates nor registers a C++ toolchain. Repository names must be globally unique
+among modules using this extension; a rule set should therefore include its
+module name in `name`.
+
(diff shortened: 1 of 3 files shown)


fmeum:
This pattern is dangerous as the prefixing isn't enforced and if used widely, this will result in many different LLVM versions being downloaded.

Instead, how about we add alias targets for the main tools under some package in the `@toolchains_llvm` repo that are then resolved to a registered LLVM toolchain? It's unlikely that a dep module would want to pin a very particular version.

@dzbarsky

helly25:
> This pattern is dangerous as the prefixing isn't enforced and if used widely, this will result in many different LLVM versions being downloaded.
> 
> Instead, how about we add alias targets for the main tools under some package in the `@toolchains_llvm` repo that are then resolved to a registered LLVM toolchain? It's unlikely that a dep module would want to pin a very particular version.
> 
> @dzbarsky

Could work, though in my current particular case I need access to all (static) lib-clang archive files.

PR #860, opened 2026-09-20
Add configurable linker selection

helly25:
## Summary

Add a general per-target linker selector. The new `linker` dictionary accepts:

- `lld` (the default), preserving the bundled and hermetic linker
- `auto`, resolving the execution platform native linker (`xcrun --find ld` on Darwin, `ld` from `PATH` on Linux)
- an absolute executable path

The selected linker path is used consistently by the Clang driver and Bazel tool paths. Linker capabilities now follow the selected implementation: bundled LLD advertises `supports_start_end_lib`, while arbitrary local linkers conservatively do not. Auto selection requires matching execution and target operating systems; explicit paths remain available for deliberately provisioned cross-linkers.

This is intentionally backend-neutral groundwork. A follow-up PR will compose an optional downloaded mold linker for ELF targets without changing the meaning of `auto`.

## Motivation

Recent macOS SDKs can contain TAPI architecture entries that older bundled `ld64.lld` versions cannot parse. Selecting the Xcode linker fixes that case, but the configuration surface should also support other linker implementations and platforms.

## Hermeticity

`auto` and absolute paths are explicitly local execution-platform dependencies. The documentation calls out the remote-execution requirement, and repository configuration tracks `PATH`, `DEVELOPER_DIR`, and `SDKROOT`. Bundled LLD remains the default.

## Tests

- Added a focused macOS/Linux CI matrix in bzlmod and WORKSPACE modes.
- bzlmod exercises `auto`; WORKSPACE exercises an explicit `/usr/bin/ld` path.
- Verified locally on macOS that the generated `CppLink` action contains the Xcode-resolved linker path and the native linker completes the repository test suite.
- `bazel test //toolchain/internal:all`
- `trunk check` on all changed files

Follow-up: optional mold integration and ELF-only validation.


fmeum (approved):

AustinSchuh:
A friend reported:

```
Tests: 9/12 passed, 3 skipped (msan/dwp/extra-files — not applicable to this platform, expected)

(with linker="auto") Confirmed via verbose linker output (-Wl,-v) that Apple's native ld (Xcode's ld-1230.1) was actually invoked — not LLVM's bundled LLD — matching the documented auto behavior (xcrun --find ld)
```

Belated, looks good!

